PR #121, opened 2015-02-10
Stop workers in additional cases when exception is raised in processes

brendar:
* Fixes case where map and each did not stop workers when a start or
  finish hook raised an exception while running in processes
* Fixes case where each did not stop workers when a worker raised an
  exception while running in processes
* Fixes some typos in map spec case comments


spec/cases/each_with_threads_and_exceptions.rb
```diff
@@ -0,0 +1,12 @@
+require './spec/cases/helper'
```
grosser: pretty much copy paste, maybe make these pairs into 1 file and pass in `ENV['THREAD']`

spec/cases/map_with_processes_and_exception_in_start.rb
```diff
@@ -0,0 +1,16 @@
+require './spec/cases/helper'
+
+begin
+  start = lambda do |_item, _index|
```
grosser: same here: `ENV['START']` -> only 1 file

lib/parallel.rb
```diff
@@ -56,7 +56,11 @@ def work(data)
       end
 
       begin
-        Marshal.load(read)
+        result = Marshal.load(read)
+        if ExceptionWrapper === result
+          raise result.exception
+        end
+        result
       rescue EOFError
```
grosser: this would make any EOFError in a block turn into DeadWorker, the original code was more like:

```Ruby
result = begin
  Marshal.load(read)
rescue EOFError
  raise DeadWorker
end
raise result.exception if ExceptionWrapper === result
result
```

lib/parallel.rb
```diff
@@ -267,20 +271,18 @@ def work_in_processes(items, options, &blk)
...
-                worker.work(items.pack(item, index))
-              end
-
-              if ExceptionWrapper === output
-                exception = output.exception
+              begin
+                results[index] = with_instrumentation item, index, options do
+                  worker.work(items.pack(item, index))
+                end
+              rescue StandardError => e
```
grosser: this will no longer bubble up the DeadWorker error, not sure why this is not causing test errors ... maybe it can go away ...

grosser:
looks like a simpler approach, I'll have to take a closer look at why basically disabling the DeadWorder logic did not break anything ... maybe it's useless

lib/parallel.rb
brendar: Missed that. Thanks. I'll rearrange it.

spec/cases/each_with_threads_and_exceptions.rb
brendar: Cool, considered DRYing these up, but I wasn't sure if adding some abstraction would be distracting. Would you be in favor of something like this?

`spec/cases/with_exceptions.rb`
```ruby
require './spec/cases/helper'

method = ENV['METHOD']
in_worker_type = "in_#{ENV['WORKER_TYPE']}"

begin
  Parallel.public_send(method, 1..100, in_worker_type => 4) do |x|
    sleep 0.1 # so all processes get started
    print x
    raise 'foo' if x == 1
    sleep 0.1 # so now no work gets queued before exception is raised
  end
rescue
  print ' raised'
end
```

lib/parallel.rb
brendar: The exception (DeadWorker or otherwise) will still get re-raised in `handle_exception`. The `it "raises DeadWorker..."` spec covers that.

spec/cases/each_with_threads_and_exceptions.rb
brendar: Invocation in the specs would remain pretty clear:
```ruby
it 'stops all workers when one fails in thread' do
  `METHOD=each WORKER_TYPE=threads ruby spec/cases/with_exceptions.rb 2>&1`.should =~ /^\d{0,4} raised$/
end
```

grosser: Looks good :)

On Thu, Feb 12, 2015 at 7:06 AM, Brendan Dougherty <notifications@github.com

brendar:
Ready for another look. Let me know if you'd like me to squash the commits.

spec/cases/with_exception_in_finish.rb
```diff
@@ -0,0 +1,19 @@
+require './spec/cases/helper'
```
grosser: could make this `with_exception_in_hook` and pass start or finish as HOOK

spec/cases/with_exception_in_finish.rb
```diff
@@ -0,0 +1,19 @@
...
+
+method = ENV['METHOD']
```
grosser: if you use `ENV.fetch('METHOD')` it blows up nicely when it's not passed

spec/parallel_spec.rb
```diff
@@ -238,19 +238,35 @@ def cpus
     end
 
     it 'stops all workers when one fails in process' do
-      `ruby spec/cases/map_with_processes_and_exceptions.rb 2>&1`.should =~ /^\d{4} raised$/
+      `METHOD=map WORKER_TYPE=processes ruby spec/cases/with_exception.rb 2>&1`.should =~ /^\d{4} raised$/
```
grosser: :heart:

grosser:
:+1: looks great, I'll merge this once I'm back home :)

spec/cases/with_exception_in_finish.rb
brendar: Cool. Will do.

spec/cases/with_exception_in_finish.rb
brendar: I kept them separate since they require different logic in both the hook and the workers to ensure consistent behavior

grosser:
thx a lot, 1.4.0 is out :)

PR #123, opened 2015-02-18
Add progress_options option to Parallel that specifies options to pass a...

danfinnie:
...long to the ruby-progressbar library.

My use case for this is outputting the progress bar on standard error by using Parallel like this:

```ruby
Parallel.map(ary, progress: "Calculating", progress_options: { output: STDERR })
```

This would replace the monkey-patch here: https://github.com/MarkyMarkMcDonald/twitch_arena_parsing/commit/ac43c073ebdf7c1e387c74f8d801714af90191d0

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -201,14 +201,20 @@ module Parallel
     private
 
     def add_progress_bar!(items, options)
-      if title = options[:progress]
+      if progress_options = options[:progress]
         raise "Progressbar and producers don't mix" if items.producer?
         require 'ruby-progressbar'
-        progress = ProgressBar.create(
-          :title => title,
-          :total => items.size,
-          :format => '%t |%E | %B | %a'
-        )
+
+        if progress_options.respond_to? :to_str
+          progress_options = { title: progress_options.to_str }
+        end
+
+        progress_options = {
+          total: items.size,
+          format: '%t |%E | %B | %a'
+        }.merge(progress_options)
+
+        progress = ProgressBar.create(progress_options)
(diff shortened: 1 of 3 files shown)


grosser:
how about using `progress: {title: 'Processing', output: STDERR}` so the other logic does not need to change ?

danfinnie:
@grosser That would solve the use case I ran into.  It seems like there are other options that people might want to pass, however.  If you just pass along a hash, then you don't have to individually support each change to the options hash that ruby-progressbar makes.

If you're against doing the hash, I can update this PR.

grosser:
merge the hash with the defaults

```Ruby
if set = options[:progress]
  set = {title: set} if set.is_a?(String)
  set = defaults.merge(set)
  ...
end
```

danfinnie:
@grosser Gotcha, I didn't fully understand what you were saying at first.  I thought that you meant to pass the whole parameter hash along to ruby-progressbar.  I switched it up to do what you suggested.

I wanted to have the passed in format and size total options to override the defaults that parallel sets (same as your example above).  Testing this was hard because ruby-progressbar ignores the :format option for non-TTYs, which is the case for the tests.  I ended up faking out the `tty?` method to get ruby-progressbar to use the passed in :format, but I'm not too happy about it.  What do you think?  Do you see a better way to test the prioritization of the ruby-progressbar options?  Is it worth it to test this?

grosser:
Looking good!

grosser:
released as 1.4.1!

danfinnie:
Thanks!

Issue #128, opened 2015-04-18
Parallel::Break and Parallel::Kill not caught when running on a platform without fork support


RJPercival:
If running on a platform without fork support (e.g. Windows), and the default options are used (i.e. run in separate processes), Parallel#work_direct will be called. This method does not catch exceptions and pass them to Parallel#handle_exception, as Parallel#work_in_threads and Parallel#work_in_processes do. This results in the exception reached user code, which is not expected behaviour.

grosser:
Yep sounds correct, can you fix it ?

PyroSpirit:
Sure, I'll send a pull request shortly. I can't easily test it though, as the RSpec tests fail dramatically on Windows.

PR #130, opened 2015-04-22
Suppress Break/Kill exceptions during #work_direct

RJPercival:
Parallel#work_direct now suppresses Parallel::Break and Parallel::Kill
exceptions, just like #work_in_threads and #work_in_processes already
did. #work_direct is called instead of #work_in_processes on platforms
where fork is not supported, e.g. Windows.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -226,12 +226,19 @@ module Parallel
 
     def work_direct(items, options, &block)
       results = []
-      items.each_with_index do |item, index|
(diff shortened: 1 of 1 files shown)


grosser:
a failing test would be great so nobody breaks it later on

ndbroadbent:
@grosser - Sorry I guess `(closes #128 and #130)` in my commit message didn't work, it needed to be `(close #128, close #130)`.

But this one can be closed too.

PR #132, opened 2015-04-25
Pass index number of process/thread to the producer lambda

jonswar:
Pass index number of process/thread to the producer lambda, so that items can be produced differently for different processes/threads.

For example, when precompiling assets in parallel, it can be more efficient for one process to handle all javascript and another to handle all css (otherwise, the processes end up duplicating a lot of dependency work).

The change should be backwards compatible with lambdas with no declared parameters.

I can add tests if desired but wanted to find out first if this change was acceptable to you.

Thanks for a great gem!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -37,6 +37,7 @@ items = [1,2,3]
 Parallel.each(lambda{ items.pop || Parallel::Stop }){|number| ... }
 ```
 
+The lambda receives a single argument, the index of the process/thread (0, 1, ...), in case you want to produce items differently for different processes/threads.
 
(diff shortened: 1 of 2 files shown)


grosser:
hmm so you are trying to make one process do similar work ... how about this ?

```Ruby
todo = [
 [js, js, js, js],
 [css,css,css,css]
]
Parallel.map(todo, in_processes: 2) { |list| do_some_work }
```

jonswar:
Ugh, that should have been obvious. Thanks! I still think there might be some value in passing the index number, but not for this case.

PR #135, opened 2015-06-13
reconnect

grosser:
parking this here until I can confirm this works ... maybe needs some tests ...

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -99,10 +99,13 @@ module Parallel
       end
     end
 
+    # generate item that is sent to workers
+    # just index is faster + less likely to blow up with unserializable errors
     def pack(item, index)
       producer? ? [item, index] : index
     end
 
+    # unpack item that is sent to workers
     def unpack(data)
       producer? ? data : [@source[data], data]
     end
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -354,6 +354,8 @@ module Parallel
           parent_write.close
           parent_read.close
 
+          active_record_reconnect(options[:active_record])
+
           process_incoming_jobs(child_read, child_write, job_factory, options, &block)
         ensure
           child_read.close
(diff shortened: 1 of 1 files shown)


seanarnold:
This didn't fix it for me :(

```
ActiveRecord::StatementInvalid: PG::UnableToSend: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

grosser:
meh :(

does this work ?
```Ruby
disconnect
Parallel do
  connect
end
```

Daniel-ltw:
```ruby
ActiveRecord::Base.connection_pool.with_connection
```

Have you guys tried something like this?

grosser:
that looks nice ... better then disconnect/reconnect for sure ... would be great to have a failing spec and then a fix that can be triggered by and option like `active_record: :reconnect`

Daniel-ltw:
Also from the looks of it, this is orm dependent and it only applies to ActiveRecord. 

If you are looking at a wider audience, maybe try supporting Sequel and ROM as well.

grosser:
let's fix the 99% usecase first :D
next important would be reconnecting redis afaik

Daniel-ltw:
The other question I have is, if this is a requirement of the gem? 

Another way to put this is, is the gem required to reconnect to the database? 
Is reconnecting to the database a responsibility of the gem?

seanarnold:
Before this goes anywhere it would be good to add some tests.

@Daniel-ltw that suggestion is already on the README. Maybe that is enough?
I agree that this shouldn't choose a single ORM, and the Gem says it's for 'Ruby' not Rails, so maybe the README section about ActiveRecord is enough?

steakknife:
Btw, having issues using parallel with sequel (`SQLite3::SQLException: no such table: {{tablename}} (Sequel::DatabaseError)` (works without parallel)

Issue #129, opened 2015-04-21
random nils from Parallel.map in jruby


kurie:
In JRuby 1.7.4 with parallel 1.4.1, I get the following behavior where calling a constructor in Parallel.map results in a small random number of nils:

```ruby
>> class Foo
>>   def initialize(a,b)
>>     @a = a
>>     @b = b
>>   end
>> end
>> range = (0...1500).to_a
>> p Parallel.map(range) { |val| Foo.new(42, val) }.count(&:nil?)
7
>> p Parallel.map(range) { |val| Foo.new(42, val) }.count(&:nil?)
0
>> p Parallel.map(range) { |val| Foo.new(42, val) }.count(&:nil?)
4
>> p Parallel.map(range) { |val| Foo.new(42, val) }.count(&:nil?)
0
>> p JRUBY_VERSION
"1.7.4"
>> p Parallel::VERSION
"1.4.1"
```

grosser:
oh boy ... I'd guess some threading issue :/

just returning `1` might already trigger it

might be able to debug this with placing random `Thread.exclusive` around some parts of the code and seeing if that fixes anything + increase the sample size

philipmw:
I just ran into this problem on JRuby 1.7.11 with parallel 1.6.0.

grosser:
plz try to reproduce and fix ... I got no idea where this is coming from ...

philipmw:
I am looking into it now.

For now, here's a minimal testcase:

    require 'parallel'

    l = Array.new(1000){|i| i}
    good = true
    while good
      pr = Parallel.map(l, {in_threads: 4}){|x| x+1}
      sr = l.map{|x| x+1}
      good = pr == sr
      print '.'
    end
    puts "\npr.compact.count=#{pr.compact.count}"
    puts "sr.compact.count=#{sr.compact.count}"

Output with JRuby 1.7.19:

    .......
    pr.compact.count=998
    sr.compact.count=1000

philipmw:
I believe I found the problem. I'll post a PR.

grosser:
fixed in 1.6.1

PR #138, opened 2015-07-21
Parallel.map writes its output safely (issue #129)

philipmw:


lib/parallel.rb
```diff
@@ -298,6 +300,7 @@ def work_in_threads(job_factory, options, &block)
     def work_in_processes(job_factory, options, &blk)
       workers = create_workers(job_factory, options, &blk)
       results = []
+      results_mutex = Mutex.new # arrays are not thread-safe
```
grosser: ` on jRuby`

grosser:
a little sad that he code needed to be added twice, but a big step forward :D

grosser:
thanks, 1.6.1 is out :D

PR #139, opened 2015-07-21
Remove Gemfile.lock since this is a gem

philipmw:
Based on this guidance:
  http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


grosser:
nope, this article is pure evil :D 

having no Gemfile.lock means:
 - randomly breaking builds on ci
 - no known good state to fall back on and do incremental updates
 - not being able to work on/fix x year old projects
 - random failures for every new contributor (gem author is fine though)

so if you care about finding new issues, make travis delete the Gemfile.lock

philipmw:
> * randomly breaking builds on ci
> * random failures for every new contributor (gem author is fine though)
> * not being able to work on/fix x year old projects

These three hypotheticals would be caused by a new release of a dependency that breaks this gem. This means that the Gemfile's dependency specifications are too loose, and the gem is now unusable by the general public!

Having a Gemfile.lock wouldn't help the situation because --- as the article points out --- that file is used only in gem development, not when the gem is part of another application.

By having a Gemfile.lock committed to git, you're setting up a scenario where all gem contributors are happy because the gem appears to work for them, while gem users (who aren't constrained by the Gemfile.lock) try the gem, find that it's broken with their dependency closure, and silently move on to something else.

This is actually already the case with` rspec-rerun`, which I address as part of this PR. The latest version of `rspec-rerun` breaks the way we include it in this gem's Rakefile. We didn't know. Fortunately this didn't break the gem for end-users because it's just specs.

> * no known good state to fall back on and do incremental updates

Travis CI's build logs help us to know the last dependency closure that built successfully.

> so if you care about finding new issues, make travis delete the Gemfile.lock

If we do this, then Travis' build environment is virtually guaranteed to be different from the developer's. This risks specs passing on the developer's machine, but failing in Travis -- and the gem developer may not know enough about Ruby development to realize the difference.

grosser:
```
These three hypotheticals would be caused by a new release of a dependency that breaks this gem. This means that the Gemfile's dependency specifications are too loose, and the gem is now unusable by the general public!
```
better lock down the patch version then ... welcome to dependency hell of not being able to update this or that gem, some loose coupling might be good like minor releases etc, but this gem does not depend on any gems, so this discussion is pointless :D 

 - I'm not going to dig through travis logs to find a working version
 - still breaks for any new contributor, making it harder to contribute

`and the gem developer may not know enough about Ruby development to realize the difference`
so it's better to fail locally and pass on travis ? or work for me but not for a contributor ... that's not any better.

for a gem that has dependencies this discussion might have some merit ... like removing the lock on travis or introducing minor version locks on dependencies, but this gem has none, so it will only make development harder, which I'm not interested in ;)

PR #141, opened 2015-07-27
Fixes #140 (travis testing for other ruby impls)

### LINKED ISSUE #140: Add JRuby to the testing matrix for travis
JRuby should be tested via Travis. Rubinus wouldn't be a bad idea either.

LightGuard:
This adds JRuby in 1.9 mode (which should track the 1.7 branch) and also JRuby 9000 which was recently released (it's ruby 2.2.2). These are run on three different JVMs as well. It should also test against Rubinus 2.

There's also no point in testing on the different jvms for straight-up mri so those combinations are excluded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,28 @@ rvm:
   - 2.0.0
   - 2.1.6
   - 2.2.2
-
+  - jruby-19mode
+  - jruby-9.0.0.0
+  - rbx-2
+jdk:
+  - openjdk7
+  - oraclejdk7
+  - oraclejdk8
+matrix:
+  exclude:
+    - rvm: 1.9.3
+      jdk: oraclejdk8
+    - rvm: 1.9.3
+      jdk: oraclejdk7
+    - rvm: 2.0.0
+      jdk: oraclejdk8
+    - rvm: 2.0.0
+      jdk: oraclejdk7
+    - rvm: 2.1.6
+      jdk: oraclejdk8
+    - rvm: 2.1.6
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,4 +6,28 @@ rvm:
   - 2.0.0
   - 2.1.6
   - 2.2.2
-
+  - jruby-19mode
+  - jruby-9.0.0.0
+  - rbx-2
+jdk:
+  - openjdk7
```
grosser: there are no java specific features / feels like a it of overkill ... how about latest jdk only ?

grosser:
they all seems to fail pretty nasty :(

.travis.yml
LightGuard: There are some invoke dynamic changes in 7 and 8. I'd suggest at least testing those.

LightGuard:
The mysql gem blows up on non-rmi rubies because it's native. I'll have to fix that. On Rubinus it looks like Parallel just doesn't work. I'll add it to the allowed failures right now until it's figured out.

.travis.yml
grosser: how about a matrix with jruby19+jdk7 and 9000+jdk8 ?

LightGuard:
It's interesting that tests are failing on jruby, do they just not work?

grosser:
yeah ... jruby was always flaky ... can you do a quick sanity check or add skips, so we get at least 1 reliably green jruby build ? :)

PR #153, opened 2016-01-19
add options[:setup_proc] and master process waits all worker being ready

taiyow:
add options[:setup_proc] to do some setup task before starting actual job.

To be sure all workers finished setup task, master process does not start actual job and progress-bar until all worker being ready.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -222,7 +222,6 @@ module Parallel
       size = [job_factory.size, size].min
 
       options[:return_results] = (options[:preserve_results] != false || !!options[:finish])
-      add_progress_bar!(job_factory, options)
 
       if size == 0
         work_direct(job_factory, options, &block)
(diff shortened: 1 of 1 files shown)


taiyow:
sorry I mistook to send.

grosser:
hmm so a setup that runs once per job ?
maybe `:job_setup => lambda { xxx }`

but something simple like this might already help ...

```
Parallel.map do
  @setup ||= do_some_foobar
  do_the_job
end
```

taiyow:
Thank you to reply.

What I want is,
* to run once per worker
* don't include setup time in progress time

With your example code, during `do_some_foobar` elapsed time of progress bar counts up.

I use parallel gem in my loadtest (it's very grate gem, thank you very much!!) and
in some cases first I connect to the server and then send request repeatedly with connection keep-alive.
I want to measure request time except connection time.

grosser:
something like this ?
```
request_time = Parallel.map do
  @setup ||= do_some_foobar
  Benchmark.realtime { do_the_job }
end.sum
```

taiyow:
Ah, yes, with using`Benchmark.realtime` and `options[:finish]` I can measure correct time for each requests.

And to run once worker, I can use your first suggestion
`@setup ||= do_some_foobar`

Thank you very much!!

PR #155, opened 2016-01-24
Feature/add begin end

taiyow:
new key `:begin` and `:end` enable to call your function before and after whole job running.


taiyow:
tested with dparallel-sample.

taiyow:
I'm very sorry I made mistook again... please ignore this.

grosser:
np, distributed parallel sounds interesting ... maybe something like that already exists ... like a map-reduce library or hadoop ...

PR #157, opened 2016-01-26
add queue interface

grosser:
Hi!
I have a parallel queue implementation ready and in use for some years (https://github.com/Try2Code/jobQueue). Your gem offers more flexibility in terms of controlling the jobs and getting results back. So I started to port some of my scripts from jobQueue towards parallel. But large blocks do not seem to fit into the array-like approach of Parallel.map. Here, the queue works better for me.

My question is, if the queue-class could be integrated into the parallel gem?

cheers
ralf

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -1,7 +1,6 @@
 require 'rbconfig'
 require 'parallel/version'
 require 'parallel/processor_count'
-require 'parallel/queue'
 
 module Parallel
   extend Parallel::ProcessorCount
@@ -236,10 +235,6 @@ module Parallel
 
     def map_with_index(array, options={}, &block)
       map(array, options.merge(:with_index => true), &block)
-    end
-
-    def queue
-      ParallelQueue.new(processor_count)
     end
 
     private
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -1,6 +1,7 @@
 require 'rbconfig'
 require 'parallel/version'
 require 'parallel/processor_count'
+require 'parallel/queue'
 
 module Parallel
   extend Parallel::ProcessorCount
@@ -235,6 +236,10 @@ module Parallel
 
     def map_with_index(array, options={}, &block)
       map(array, options.merge(:with_index => true), &block)
+    end
+
+    def queue
+      ParallelQueue.new(processor_count)
     end
 
     private
--- a/lib/parallel/queue.rb
+++ b/lib/parallel/queue.rb
@@ -1,56 +1,32 @@
+require 'thread'
 # ==============================================================================
-# Author: Ralf Mueller, ralf.mueller@mpimet.de                                 
+# Author: Ralf Mueller, ralf.mueller@mpimet.mpg.de
 #         suggestions from Robert Klemme (https://www.ruby-forum.com/topic/68001#86298)
-#                                                                            
+#
 # ==============================================================================
-# Sized Queue for limiting the number of parallel jobs                       
-# ==============================================================================                          
-class ParallelQueue
-  include Parallel::ProcessorCount
+include Parallel::ProcessorCount
 
-  attr_reader :workers, :threads
-
-  # Create a new queue qith a given number of worker threads
-  def initialize(nWorkers=processor_count,debug=:off)
-    @workers = nWorkers
-    @queue   = Queue.new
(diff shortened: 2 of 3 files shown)


grosser:
not sure how this is supposed to be used / how it is different then the current queue usage ...

```Ruby
q=Queue.new
q.push 1
q.push Parallel::Stop
Parallel.each(q, in_threads: 2){ |x| puts x  }
```

Try2Code:
I normaly use JobQueue like this:
https://code.zmaw.de/projects/cdo/wiki/Cdo%7Brbpy%7D#Multi-threadding
https://github.com/Try2Code/andAllTheRestOfIt/blob/master/building/Rakefile#L142
https://github.com/Try2Code/codegraph/blob/master/lib/codegraph.rb#L54

I wasn't aware of the possibility of letting Parallel run through the queue. but that's quite smart. Queue works like an array, so it fits to the each/map interface. Maybe I can get the results of each block with manual mutex'ing. I think I can get rid of the class and just keep the run-method for yielding the code blocks. I will test it.

Thanks for the hint!

Try2Code:
I checked some implementations and ended up with an extension on the Queue class. It can be activated by 

   require 'parallel/queue'

the real benefit is the ability to collect procs and blocks for parallel execution. I don't see any other possibility to do this. 
What do you think? Is it worth to include it?

grosser:
I don't understand what this would be used for / what an example usage would look like.
2 red flags for me:
 - global includes
 - modifications of the existing Queue class

Try2Code:
I changed from class extension to a child class of Queue, shifted the include statement and added some unit tests (sorry, I didn't use rspec, but I'm not familiar with it)

Maybe I should have started with _why_ I'd like to have that integrated - my bad:
Basically I'd like to get rid of my JobQueue library and switch over to yours, because its is more powerful and general (support threads _and_ processes, getting results). The only thing, I'm missing is the ability to call arbitrary blocks in parallel. I use this kind of parallelism in analysing/visualisation of climate model data. Here I have different kinds of input data (e.g. gridded data, profile data, satellite data, text) which need different routines for generating plots out of it. These kinds of calls are totally independent of each other, so they should be called in parallel. the [_test_blocks_  method](https://github.com/Try2Code/parallel/commit/133e7c31f0a753779ad5c09166d3516d0e231f81#diff-3d3049faf443986d46c88b8da670f9f6) is a very simple example of this.

grosser:
hmmm so what you are trying to do is push a block into a queue and have them evaluated in parallel ?

```Ruby
lambdas = {
  a: lambda { |x| "a" * x },
  b: lambda { |x| "b" * x }
}
q = Queue.new

[1,2,3].each do |n|
  q.push [:a, n]
  q.push [:b, n]
end

q.push Parallel::Stop

puts Parallel.map(q, in_threads: 2) { |l,n| lambdas[l].call(n) }
```

Try2Code:
vs.

```ruby
require 'parallel/queue'
q  = ParallelQueue.new
    
[1,2,3].each do |n|
  q.push {
    "a" * n
  }
  q.push {
    "b" * n
  }
end
puts q.run
```
your decision :smirk:

grosser:
the point of the queue interface is that you can fill it forever and when you are done you add the stop,
if you already know all the values up front, then an array would work equally well ?
then you can simply put lambdas into an array:

```Ruby
Parallel.map([lambda{ 1 }]){ |x|  x.call }
```

Try2Code:
Maybe that's just a matter of taste: You prefer lambdas, I prefer blocks. 
I can understand, that you're hesitant, if the problem can be solved with lambdas out of the box. 

If you don't want to take over the code - no problem. I can still write my library as a small extension of yours. 

have a nice weekend!
ralf

grosser:
Since you are not using the Queue correctly you might as well use an array:

```
Parallel.map([lambda{ 1 }]) { |x| x.call }
```

using a queue with lambdas as ParallelQueue does, does not work with processes:

```
q = Queue.new
q.push lambda { 1 }
Parallel.map(q) { |x| x.call }
TypeError: no _dump_data is defined for class Proc
```

and using directly with lambdas is pretty straight forward, so less apis/things to learn

```
q = Queue.new
q.push lambda { 1 }
q.push Parallel::Stop
Parallel.map(q, in_threads: 2) { |x| x.call }
```

PR #161, opened 2016-02-22
Don't call :finish hook if there was an exception

ajaska:
I couldn't find any good reason for the :finish hook to be inside an ensure block.

But because it is, the :finish hook will be called with a `nil` value for `result` if there is an exception on the :start hook or `yield` call. This is (IMHO) unexpected behavior, and I was quite confused as to why my function (which should always be called with a non-nil value) was receiving `nil`.

### DIFF
--- a/spec/cases/with_break_before_finish.rb
+++ b/spec/cases/with_break_before_finish.rb
@@ -3,18 +3,17 @@ require './spec/cases/helper'
 method = ENV.fetch('METHOD')
 in_worker_type = "in_#{ENV.fetch('WORKER_TYPE')}".to_sym
 
-begin
-  finish = lambda do |_item, _index, _result|
-    print " called"
-  end
-
-  Parallel.public_send(method, 1..10, in_worker_type => 4, finish: finish) do |x|
-    sleep 0.2
-    if x != 3
-      raise Parallel::Break
-    end
-    print x
-    x
-  end
-rescue
+finish = lambda do |_item, _index, _result|
+  sleep 0.1
+  print "finish hook called"
 end
+
+result = Parallel.public_send(method, 1..100, in_worker_type => 4, finish: finish) do |x|
+  sleep 0.1 # let workers start
+  raise Parallel::Break if x == 1
+  sleep 0.2
+  print x
+  x
+end
+
+print " Parallel::Break raised"
--- a/spec/cases/with_break_before_finish.rb
+++ b/spec/cases/with_break_before_finish.rb
@@ -1,0 +1,20 @@
+require './spec/cases/helper'
+
+method = ENV.fetch('METHOD')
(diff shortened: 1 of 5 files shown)


grosser:
I think this is a feature without tests ... it is supposed to work when using `raise Parallel::Break` ... can you add a test to see that this still works ?

ajaska:
Hmm, I'm not sure I understand. When you `raise Parallel::Break`, do you still want the finish hook to be called for that worker? From what I can tell, I'm not sure that that would be useful as

- You won't have a usable result from that worker
- You are terminating early, so progress information isn't useful

I suppose it might be worth doing for compatibility reasons?

ajaska:
I went ahead and pushed up test cases for my interpretation, let me know if you think `Parallel::Break` should be a special case, and I'd be happy to push up that version.

spec/cases/with_exception_in_start_before_finish.rb
```diff
@@ -0,0 +1,23 @@
+require './spec/cases/helper'
+
+method = ENV.fetch('METHOD')
+in_worker_type = "in_#{ENV.fetch('WORKER_TYPE')}".to_sym
+
+begin
+  start = lambda do |_item, _index|
```
grosser: _xx means it's unused ... but here it is used :)

spec/cases/with_break_before_finish.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  Parallel.public_send(method, 1..10, in_worker_type => 4, finish: finish) do |x|
+    sleep 0.2
+    if x != 3
+      raise Parallel::Break
+    end
+    print x
+    x
+  end
+rescue
```
grosser: the rescue should not be necessary

spec/cases/with_exception_before_finish.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  Parallel.public_send(method, 1..10, in_worker_type => 4, finish: finish) do |x|
+    if x != 3
+      sleep 0.2
+      raise 'foo'
+    end
+    print x
+    x
+  end
+rescue
```
grosser: maybe also raise something specific and then rescue only that ... to make sure we are not raising random errors

spec/cases/with_break_before_finish.rb
grosser: the expected result would be finish getting called with 1,2,3 since these items finished ?

ajaska: ah, you're right, I structured this testcase poorly

ajaska:
Thanks for the feedback, I went ahead and made those changes. Let me know about other changes/if you want me to rebase or something.

grosser:
nice tests! :)

grosser:
1.6.2 is out!

PR #163, opened 2016-03-17
add method tasks

debbbbie:
Use `tasks` for some different works
 
 ```Ruby
 Parallel.tasks() do |tasks|
   tasks.add{puts 'one task'}
   tasks.add{system("echo 'another task'")}
 end
 ```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -37,6 +37,14 @@ items = [1,2,3]
 Parallel.each(lambda{ items.pop || Parallel::Stop }){|number| ... }
 ```
 
+Use `tasks` for some different works
+
+```Ruby
+Parallel.tasks() do |tasks|
+  tasks.add{puts 'one task'}
+  tasks.add{system("echo 'another task'")}
+end
+```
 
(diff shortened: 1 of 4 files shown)


grosser:
same as `Parallel.each([->{ puts 1 }])(&:call)` ?

debbbbie:
@grosser Yes, but more legible, isn't it?

For there are many long command in production. And priority uses will be fine, such as `jQuery`.

grosser:
I prefer simple/reusable ... already got 2 different apis and a bunch of ways to use them ...
`.each` seems more intuitive to me / more obvious what is going on / know it will be unordered ... and it's easy to switch to map/threads/processes/progress-bar etc too :)

```Ruby
Parallel.each([
   -> { puts 'one task' }
   -> { system("echo 'another task'") }
])(&:call)
```

sorry to close this, I can see you put a lot of work into this ... but I'd rather not have another api that does exactly the same with a slightly different syntax :(

debbbbie:
OK :+1:

PR #177, opened 2016-06-01
expose worker/slot number to tasks via Thread#[]

seanwalbran:
When processing tasks, set a thread local value `:parallel_worker_number` to allow the task logic to know what 'slot' it is in, so that parallel tasks can coordinate access to external resources like database schemas -- basically for the same purposes as the TEST_ENV_NUMBER environment variable in the [parallel_tests](https://github.com/grosser/parallel_tests) gem.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -277,6 +277,7 @@ module Parallel
     end
 
     def work_direct(job_factory, options, &block)
+      Thread.current[:parallel_worker_number] = 0
       results = []
       while set = job_factory.next
         item, index = set
@@ -285,6 +286,8 @@ module Parallel
         end
       end
       results
+    ensure
+      Thread.current[:parallel_worker_number] = nil
     end
 
     def work_in_threads(job_factory, options, &block)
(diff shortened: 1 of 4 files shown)


grosser:
that looks kind of useful ... how about `Parallel.worker_number` and `Parallel.worker_number=` so we are not accessing Thread state all over the place ...

seanwalbran:
@grosser Good call, how does b6965b3 look to you?

grosser:
1.9 tests seems to break ? `       No such file or directory - METHOD=each WORKER_TYPE=processes ruby spec/cases/with_worker_number.rb
`

seanwalbran:
Thanks, looks like 5ca7f2e (adding stderr redirection, like in the neighboring specs) fixes the 1.9.3-specific errors.

grosser:
1.9.0

can you add some Readme docs too ?

seanwalbran:
Sure thing, here's a PR for the docs: https://github.com/grosser/parallel/pull/178
Thanks for the merge!

PR #178, opened 2016-06-03
add Readme docs for .worker_number

seanwalbran:
...as requested here: https://github.com/grosser/parallel/pull/177#issuecomment-223657569


grosser:
But why did you break the build ? :D 

thx ❤️

PR #179, opened 2016-06-14
add option[:exit_bang_status] so child does exit!() when ending

cxhartmann:
Mitigates child workers from closing inherited connections upon exit. In particular for mysql connections owned by the parent rails app would be inadvertently closed by the child when exiting normally and result in this error for subsequent queries in the parent: "ActiveRecord::StatementInvalid: Mysql2::Error: MySQL server has gone away"

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -405,7 +405,9 @@ module Parallel
         ensure
           child_read.close
           child_write.close
+          exit!(options[:exit_bang_status]) unless options[:exit_bang_status].nil?
         end
+
       end
 
       child_read.close


grosser:
that seems like a sledgehammer approach to the problem ... most likely causes unrelated fallout like tempfiles not being closed/ensure blocks not being run etc ...
how about close + open in fork + close + open after done ... also avoid using the same connection in parallel which causes other kinds of funky issues ...

cxhartmann:
Thanks for the quick response.
> how about close + open in fork + close + open after done

I'd like to avoid breaking an uncommitted mysql transaction with that first close. In my scenario the child doesn't need to query the database, so I suppose this could be solved without exit!() if there were a way to simply unreference (without closing) the mysql connections in the child block. Do you happen to know how ?

grosser:
no idea ... dig into mysql gem or maybe activereocrd to find out ... but
still reusing the same connection that the parent uses is a bad idea ... so
better avoid it ?

cxhartmann:
Alright after second thought with the concerns you raise I agree this isn't the best approach. For my specific case, as an alternative, I reach the same outcome without a new options parameter by just defining an at_exit handler in the child block to exit!() as such: at_exit { exit!(0) } . This proceeds the exit handler the nukes mysql connections owned by the parent - still has the concerning characteristics of tempfile orphaning and ensure but in my specific case neither of those apply. Ideally removing the references to the ActiveRecord mysql connections in the child would be best, but my initial attempts to do that came up short.

PR #187, opened 2016-10-17
Inroduce the reduce method for deduplication index building and others 

simcha:
Reduce will return results colleced from each worker. So you need to merge them on the end.

``` Ruby
result = Parallel.reduce(['a','b','c','d','a','b','c','d']) do |result,x|
  result ||= Set.new
  result << x
  result
end
result.compact.reduce(&:+).sort #-> ['a','b','c','d']
```

Remember that if set is small and hi number of workers some workers can return nil. So it is better to use "compact" before merging resuls.


simcha:
I got no errors on my local run and travis got different errors on my copy of same tree:
https://travis-ci.org/simcha/parallel/builds/168296730.
It seams to be error related to process_diff method
And especially command "ps uaxw|grep ruby|wc -l"
It is not checking only our processes and sometime can result with error.

grosser:
why not map and then reduce after ?

simcha:
My case is word count. I do have 7 million files to count 5 letter words in them. So I need to feed a file and get word count hash. I could map first and have 7 million hashes but merging them is expensive operation and also memory would end. Chunking into smaller packs is not a case as merging ~4 mln. elements hash is as I said expensive. Using the parallel lib with my modification reduced the time of operation of my script 10 fold with very simple change in the code (and 16 CPUs). For that I am very grateful to you for writing the library. Maybe there is some way to do it without a "reduce" method but I do not see it. So answering your question with a question. I do map first but "why do not reduce in parallel" :)

Readme.md
```diff
@@ -37,6 +37,26 @@ items = [1,2,3]
...
 ```
 
+Reduce will return results colleced from each worker. So you need to merge them on the end.
+```Ruby
+result = Parallel.reduce(['a','b','c','d','a','b','c','d']) do |result,x|
+  result ||= Set.new
+  result << x
+  result
+end
+result.compact.reduce(&:+)
```
grosser: why does it need compact ?
... maybe do `'x' + 'p'` to show that processing was done

Readme.md
```diff
@@ -37,6 +37,26 @@ items = [1,2,3]
 Parallel.each( -> { items.pop || Parallel::Stop }) { |number| ... }
 ```
 
+Reduce will return results colleced from each worker. So you need to merge them on the end.
```
grosser: Maybe write a bit about what you just wrote in the comment ...

```
To avoid overhead from reducing big result sets at the end, use `.reduce` to bla bla
```

grosser:
hmmm would about implementing this with a finish hook ?

```Ruby
>> all = 0; Parallel.each([1,2,3,4,5], finish: -> (_,_,result) { all += result }){ |x| x + 10 }
=> [1, 2, 3, 4, 5]
>> all
=> 65
```

Readme.md
simcha: I am not absolutely sure I do understand your question but compact is only for a case when processing was not done in some process and it has nil in result. The part of question I do not understand is `'x' + 'p'` part, where to do it?

simcha:
With finish hook how is this parallel? If it runs on each worker finish it will block (?) and it will not give us the overhead reduction if we targeting the case where we replace `all += result` from your example with something longer to run. If it is run in separate workers how you make it so that it returns 1 result and not array of results?

But the finish hook gave me some idea. Now we need to wait for workers to finish to get the result and with the finish hook we can chain calculations. So we can have infinite queue and some time or count parameter to something like finish hook. Every 1 minute or every 100 inputs call the hook with results on input. Seams to me as very simplistic way of creating daemons/servers, what you think?

grosser:
The finish hook is not parallel, but ideally combining results is
fast/simple.

Yep, thats a neat way of doing a parallel daemon or get intermediate
results :)

Readme.md
simcha: That is fixed now we do not need it.

Readme.md
simcha: So I added comment similar to that

lib/parallel.rb
```diff
@@ -50,8 +50,14 @@ def initialize(read, write, pid)
     end
 
     def stop
-      close_pipes
+      write.close unless write.closed?
```
grosser: closing multiple times is not an issue, so just do it ?

lib/parallel.rb
```diff
@@ -480,10 +502,13 @@ def extract_count_from_options(options)
       [count, options]
     end
 
-    def call_with_index(item, index, options, &block)
+    def call_with_index(item, index, options, end_result, &block)
```
grosser: arguments after options are weird 🤷‍♂️

spec/cases/helper.rb
```diff
@@ -2,7 +2,7 @@
 require 'parallel'
 
 def process_diff
-  cmd = "ps uaxw|grep ruby|wc -l"
+  cmd = "ps uxw|grep ruby|wc -l"
```
grosser: why is this changed ?
can this be it's own PR with a small description ?

Readme.md
```diff
@@ -53,8 +53,27 @@ Parallel.all?([1,2,nil,4,5]) { |number| number != nil }
 # => false
 ```
 
+To avoid overhead from reducing big result sets at the end of processing, use `reduce` this method will return results collected from each worker. You need to merge them on the end to get the fully reduced result.
 
-Processes/Threads are workers, they grab the next piece of work when they finish.
```
grosser: only document 1 way of doing things, this is already complicated enough and will only be needed by very few users

grosser:
so basically what this is doing is providing a way to aggregate information per worker without running things on the main thread ?

spec/cases/helper.rb
grosser: ah found it above.

Readme.md
grosser: naming it `reduce` suggests that it will return 1 result, but it does not, also we might later add an actual `reduce`
so reduce_by_worker might be more clear

grosser:
also blow up when `isolation` option is used since these 2 make no sense together

grosser:
... also this whole thing can be implemented with some creative hacking too ...

```Ruby
sum = [nil, 0]
Parallel.map [1,2,3,4,5,6,7,8,9], in_processes: 2 do |i|
  sum[0] ||= Process.pid
  sum[1] += i
  sum
end.reverse!.uniq(&:first).map(&:last).sum

```

Readme.md
simcha: Sure

simcha:
Creative hacking way seam nice

PR #188, opened 2016-11-03
Adds comment to readme for trapping TERM

rkm3:
I spent a while trying to figure out how to use Parallel with `runit`, specifically handling the way `runit` sends the `TERM` signal. I ended up finding this option in the source, so I thought I'd leave it in the readme for the next person. 

I see the TODO at the bottom of the Readme, I wish I had the foo to implement it. Maybe next time.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -92,7 +92,7 @@ Only use if whatever is executing in the sub-command is safe to kill at any poin
 ```Ruby
 Parallel.map([1,2,3]) do |x|
   raise Parallel::Kill if x == 1# -> stop all sub-processes, killing them instantly
-  sleep 100
+  sleep 100 # Do stuff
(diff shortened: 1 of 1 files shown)


grosser:
so what happened was that the process got killed, but parallel did not clean up ?
... kind of a bug too :(

rkm3:
Yeah. Ctrl-c would kill but killing with vanilla kill would leave orphan
processes. Switch the message would reverse the behavior.

Raising Parallel::Break or Kill didn't seem to stop the subprocesses.

I'm new to the library so I figured I was doing something wrong.

grosser:
sounds like a bug ... I guess not many people use it with runit :D
... should ideally kill on TERM + Break / Kill

PR #191, opened 2017-01-14
Kill all child processes when the main process or thread is killed

nolanpro:
Hey @grosser here's the changes I made to allow parallel to kill all child processes. Let me know what you think. Thanks!

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -79,6 +79,7 @@ module Parallel
     def wait
       Process.wait(pid)
     rescue Interrupt
+    rescue Errno::ECHILD
       # process died
     end
   end
@@ -168,7 +169,9 @@ module Parallel
 
       def kill(thing)
         Process.kill(:KILL, thing)
+        Process.waitpid(thing)
       rescue Errno::ESRCH
+      rescue Errno::ECHILD
         # some linux systems already automatically killed the children at this point
         # so we just ignore them not being there
       end
@@ -198,10 +201,19 @@ module Parallel
 
   class << self
     def in_threads(options={:count => 2})
-      count, _ = extract_count_from_options(options)
-      Array.new(count).each_with_index.map do |_, i|
-        Thread.new { yield(i) }
-      end.map!(&:value)
+      begin
+        count, _ = extract_count_from_options(options)
+        threads = Array.new(count).each_with_index.map do |_, i|
+          Thread.new do
+            Thread.current.abort_on_exception = true
+            yield(i)
+          end
+        end
+        threads.map!(&:value)
+      rescue Parallel::Kill => e
+        kill_child_processes
+        kill_threads(threads)
+      end
     end
 
     def in_processes(options = {}, &block)
@@ -322,6 +334,8 @@ module Parallel
               call_with_index(item, index, options, &block)
             end
             results_mutex.synchronize { results[index] = result }
+          rescue Parallel::Kill => e
+            raise e
           rescue StandardError => e
             exception = e
           end
@@ -442,6 +456,24 @@ module Parallel
       return nil if [Parallel::Break, Parallel::Kill].include? exception.class
       raise exception if exception
       results
+    end
+
+    # kill all children of this process
+    def kill_child_processes
+      child_processes(Process.pid).each do |pid|
+        UserInterruptHandler.kill(pid)
+      end
+    end
+
+    # kill threads asynchronously (using new threads)
+    def kill_threads(threads)
+      threads.reject(&:nil?).each(&:kill)
+    end
+
+    # get child pids ordered by youngest descendants first
+    def child_processes(pid)
+      pids = `pgrep -P #{pid}`.split("\n").map(&:to_i)
+      pids.map { |p| child_processes(p) }.flatten + pids
     end
 
     # options is either a Integer or a Hash with :count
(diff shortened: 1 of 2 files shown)


spec/parallel_spec.rb
```diff
@@ -367,11 +367,39 @@ def cpus
...
+      Parallel.map([1,2], :in_threads => 2) do |i|
+        if i == 1
+          sleep(0.1)
+          raise Parallel::Kill
+        else
+          p = IO.popen('sleep 5; echo "Should not get here"')
+          pid = p.pid
+          p.read
+        end
+      end
```
grosser: should inspect the result here and assert it was fast / does not include "Should not get here"

spec/parallel_spec.rb
```diff
@@ -367,11 +367,39 @@ def cpus
...
+      `ps -e | grep '^#{pid}'`.should == ''
```
grosser: should grep for a ` ` after to not match 1234 with 123

spec/parallel_spec.rb
```diff
@@ -367,11 +367,39 @@ def cpus
...
+      result = Parallel.map([1,2], in_threads: 2) do |i|
+        if i == 1
+          sleep(0.1)
+          raise Parallel::Kill
+        else
+          sleep(3)
+          "Should not get here"
+        end
+      end
+      result.should_not include("Should not get here")
```
grosser: could do a more precise `should == nil` ... or whatever the result would be

spec/parallel_spec.rb
```diff
@@ -367,11 +367,39 @@ def cpus
       `ruby spec/cases/eof_in_process.rb 2>&1`.should include 'Yep, EOF'
     end
 
-    it "can be killed instantly" do
+    it "processes can be killed instantly" do
```
grosser: weird english ... `it can be killed instantly in process mode` ?

spec/parallel_spec.rb
```diff
@@ -194,7 +194,7 @@ def cpus
     end
 
     it "raises when a thread raises" do
-      lambda{ Parallel.in_threads(2){|i| raise "TEST"} }.should raise_error("TEST")
+      lambda{ Parallel.in_threads(2){|i| raise "TEST" if i == 1} }.should raise_error("TEST")
```
grosser: why is this necessary ... it should also work when both raise ...

lib/parallel.rb
```diff
@@ -444,6 +458,24 @@ def handle_exception(exception, results)
       results
     end
 
+    # kill all children of this process
+    def kill_child_processes
+      child_processes(Process.pid).each do |pid|
+        UserInterruptHandler.kill(pid)
```
grosser: how about `UserInterruptHandler.kill(child_processes(Process.pid))` or even automatic child killing via `UserInterruptHandler.kill pid`

lib/parallel.rb
```diff
@@ -444,6 +458,24 @@ def handle_exception(exception, results)
...
+      end
+    end
+
+    # kill threads asynchronously (using new threads)
+    def kill_threads(threads)
+      threads.reject(&:nil?).each(&:kill)
```
grosser: `.reject(&:nil?)` == `.compact`

lib/parallel.rb
```diff
@@ -444,6 +458,24 @@ def handle_exception(exception, results)
...
+    end
+
+    # get child pids ordered by youngest descendants first
+    def child_processes(pid)
+      pids = `pgrep -P #{pid}`.split("\n").map(&:to_i)
+      pids.map { |p| child_processes(p) }.flatten + pids
```
grosser: `.flat_map` == `.map{}.flatten`

lib/parallel.rb
```diff
@@ -168,7 +169,9 @@ def kill_on_ctrl_c(pids, options)
 
       def kill(thing)
         Process.kill(:KILL, thing)
+        Process.waitpid(thing)
       rescue Errno::ESRCH
```
grosser: can be on same line `rescue X, Y`

lib/parallel.rb
```diff
@@ -79,6 +79,7 @@ def work(data)
     def wait
       Process.wait(pid)
     rescue Interrupt
+    rescue Errno::ECHILD
```
grosser: same line here too ?

grosser: actually a good sign that you found this bug :D

grosser:
Looks really good, just a few nitpicks here and there :D

spec/parallel_spec.rb
nolanpro: I added this because ```abort_on_exception = true``` was causing problems when two exceptions were thrown at the exact same time. Sometimes only one was getting rescued. I'm not sure if this is a bug in ruby threads or if it's a synchronization issue. But ether way, now that I think about it, we should account for that.

I might have found a better alternative to abort_on_exception so I'm going to play with that in the update to the PR

grosser:
Alternatively do a short sleep ... But I think it's realistic to have two
threads raise at the same time, so would be nice if we supported that

xjose97x:
Why hasn't this PR been merged yet?

grosser:
Would at least need a fresh PR since lots has changed since this was opened.

Issue #198, opened 2017-03-10
Would be really handy to have more methods: `flat_map`, `reject`, `find`, `any?`, `all?`


ndbroadbent:
I was also thinking about `reduce`, but I guess you couldn't share the accumulator variable across processes.

I've been trying to figure out how to implement `any?` and `all?` using `raise Parallel::Kill`, but it doesn't seem very efficient. Is there a better way to do the following?

```
        result = Parallel.map(array) do |item|
          unless condition(item)
            raise Parallel::Kill
          end
        end

        if result.nil?
          # at least one condition was false
        else
          # all conditions were true
        end
```

I don't really care about the result array (in this example it's just full of nils).

grosser:
that looks about right ...
I think it would be neat to have them ... maybe use Break instead to avoid
killing things mid-process

ndbroadbent:
Thanks for your reply! I've added a pull request for `any?` and `all?`. I'll have a go at some of the other ones, too.

PR #199, opened 2017-03-10
Added any? and all? methods

ndbroadbent:
See: #198

I think these could be very useful.


lib/parallel.rb
```diff
@@ -215,6 +215,14 @@ def each(array, options={}, &block)
       array
     end
 
+    def any?(array, options={}, &block)
+      map(array, options.merge(:break_condition => :any), &block || -> (v) { v })
+    end
+
+    def all?(array, options={}, &block)
+      map(array, options.merge(:break_condition => :all), &block || -> (v) { v })
```
grosser: return `!all(*args)` here ?

grosser:
lots of code for a simple change ... I was hoping for something that did not require changing the internals like

```
def any?(*args, &block)
  !each(*args) { raise Parallel::Kill if block.call }
end
```

ndbroadbent:
Thanks for the feedback, I agree this is not very elegant. I'll try to trim it down to something more reasonable

ndbroadbent:
Updated!

Haha you're right, this is WAY better. I just needed to fix that `work_direct` issue in #201

Readme.md
```diff
@@ -37,6 +37,18 @@ items = [1,2,3]
 Parallel.each( -> { items.pop || Parallel::Stop }) { |number| ... }
 ```
 
+You can also call `any?` or `all?`, which work the same way as `Array#any?` and `Array#all?`.
+
+```Ruby
+items = [1,2,3,4,5,6,7]
+Parallel.any?(items) { |number| number == 4 }
```
grosser: save a line by inlining the `items=` ?

lib/parallel.rb
```diff
@@ -215,6 +215,20 @@ def each(array, options={}, &block)
       array
     end
 
+    def any?(array, options={}, &block)
+      block ||= -> (v) { v }
+      !map(array, options.merge(:preserve_results => false)) do |*args|
```
grosser: can use `each` ?

lib/parallel.rb
```diff
@@ -215,6 +215,20 @@ def each(array, options={}, &block)
...
+        raise Parallel::Kill if block.call(*args)
+      end
+    end
+
+    def all?(array, options={}, &block)
+      block ||= -> (v) { v }
```
grosser: better raise when there is no block ?

grosser:
lookin good!

lib/parallel.rb
ndbroadbent: `#each` returns the original array, so I couldn't do that, I need the result of the `map` call

Readme.md
ndbroadbent: 👍

lib/parallel.rb
ndbroadbent: Ah, the implicit block is the default behavior in Ruby: http://apidock.com/ruby/Enumerable/any%3F

So you can call:

```
[ nil, true, 99 ].any?                            #=> true
```

grosser:
yeah, but kind of weird/useless when you call parallel without a block ?

On Fri, Mar 10, 2017 at 12:34 PM, Nathan Broadbent <notifications@github.com

grosser:
I guess that's another bug then ... .each should return nil on break:

```
[1,2,3].each{ break }
=> nil
```

... but can be it's own PR if it's not trivial to fix

On Fri, Mar 10, 2017 at 12:32 PM, Nathan Broadbent <notifications@github.com

ndbroadbent:
I pushed a fix for the each issue, so now I'm using each. Super clean!

grosser:
❤️  waiting for green :)

grosser:
1.11.0

ndbroadbent:
Awesome, thanks very much!

PR #200, opened 2017-03-10
Added find/detect, find_all/select, and reject methods

ndbroadbent:
Added some more methods from Array. The code is getting a bit hairy, so I would like to refactor it a bit and reduce duplication. Just pushing now to get some feedback.

(This PR also includes #199, to avoid dealing with merge conflicts.)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -37,7 +37,7 @@ items = [1,2,3]
 Parallel.each( -> { items.pop || Parallel::Stop }) { |number| ... }
 ```
 
-You can also call `any?` or `all?`, which work the same way as `Array#any?` and `Array#all?`.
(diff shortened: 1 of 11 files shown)


grosser:
not a big fan of this all going so deep into the code ... also not sure about find_all/reduce being that useful since you'll have to map anyway ?
... any/all at least have the benefit of better performance

ndbroadbent:
That's a good point about having to map anyway, although I think it could still be useful if you want to run an expensive computation on all your CPUs to determine which items to keep.

But I agree, this could be done a lot better. I'm about to update the `any` and `all` PR with your suggestions, so will revisit this afterwards

PR #201, opened 2017-03-10
Handle Break/Kill exceptions during #work_direct

ndbroadbent:
#130 plus a failing test

(needed this to make `any?` and `all?` much easier to implement)


lib/parallel.rb
```diff
@@ -295,13 +295,18 @@ def add_progress_bar!(job_factory, options)
...
-          call_with_index(item, index, options, &block)
+      exception = nil
+      begin
+        while set = job_factory.next
+          item, index = set
+          results << with_instrumentation(item, index, options) do
+            call_with_index(item, index, options, &block)
+          end
         end
+      rescue StandardError => e
```
grosser: StandardError is the default ... so this should be the same:

```
rescue
  exception = $!
end
```

grosser:
👍  super good ... these nasty little bugs :D

lib/parallel.rb
ndbroadbent: Updated

ndbroadbent:
I'll close this one since it's already in #199

PR #202, opened 2017-03-11
Fixes crash on Kill and Break when the better_errors gem is being used

ndbroadbent:
I use the [better_errors](https://github.com/charliesome/better_errors) gem with Rails, which ends up [making exceptions undumpable](https://github.com/charliesome/better_errors/issues/254). This includes `Parallel::Break` and `Parallel::Kill`. When debugging `ExceptionWrapper#initialize` in the Rails console, I was seeing this error:

```Ruby
TypeError - no _dump_data is defined for class Binding:
```

This commit changes `ExceptionWrapper` to return empty `Break` or `Kill` exceptions (by calling `exception.class.new`. It fixes the undumpable exception issue, and also provides the tiniest of performance improvements, because we don't need to marshall the backtrace.

(I'll push two commits separately. First the failing test, then the fix.)

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -15,13 +15,10 @@ module Parallel
   end
 
   class UndumpableException < StandardError
+    attr_reader :backtrace
     def initialize(original)
       super "#{original.class}: #{original.message}"
-      @bracktrace = original.backtrace
-    end
-
-    def backtrace
-      @bracktrace
+      @backtrace = original.backtrace
     end
   end
 
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -27,6 +27,12 @@ module Parallel
   class ExceptionWrapper
     attr_reader :exception
     def initialize(exception)
+      # Remove the bindings stack added by the better_errors gem,
+      # because it cannot be marshalled
+      if exception.instance_variable_defined? :@__better_errors_bindings_stack
+        exception.send :remove_instance_variable, :@__better_errors_bindings_stack
+      end
+
       @exception =
         begin
           Marshal.dump(exception) && exception
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -69,6 +69,3 @@ DEPENDENCIES
   rspec-rerun
   ruby-progressbar
   sqlite3
-
-BUNDLED WITH
-   1.13.6
--- a/spec/cases/parallel_break_better_errors.rb
+++ b/spec/cases/parallel_break_better_errors.rb
@@ -1,0 +1,21 @@
+require './spec/cases/helper'
+require 'stringio'
+
+class MyException < StandardError
+  def initialize(object)
+    @object = object
+  end
+end
+
+begin
+  Parallel.in_processes(2) do
+    ex = Parallel::Break.new
+    # better_errors sets an instance variable that contains an array of bindings.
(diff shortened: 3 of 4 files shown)


ndbroadbent:
Confirming that it fixes the crash in rails console:

### Before:

```
$ rails console
Loading development environment (Rails 4.0.3)
2.1.5 :001 > Parallel.each([1,2,3,4,5,6,7,8,9]) {|x| raise Parallel::Kill if x == 4 }
Parallel::UndumpableException: Parallel::Kill: Parallel::Kill
from (irb):1:in `block in irb_binding'
from /Users/ndbroadbent/.rvm/gems/ruby-2.1.5/gems/parallel-1.11.0/lib/parallel.rb:481:in `call'
```

### After:

```
$ rails console
Loading development environment (Rails 4.0.3)
2.1.5 :001 > Parallel.each([1,2,3,4,5,6,7,8,9]) {|x| raise Parallel::Kill if x == 4 }
 => nil
```

lib/parallel.rb
```diff
@@ -31,10 +28,14 @@ class ExceptionWrapper
     attr_reader :exception
     def initialize(exception)
       @exception =
-        begin
-          Marshal.dump(exception) && exception
-        rescue
-          UndumpableException.new(exception)
+        if [Parallel::Break, Parallel::Kill].include? exception.class
+          exception.class.new
```
grosser: isn't this losing the backtrace ?

grosser:
won't all exceptions be undumpable ? ... why are ours special ?

lib/parallel.rb
ndbroadbent: Yes, I was thinking that we don't use it, and these exceptions are never visible to the developer

ndbroadbent:
I figured out what better errors was doing, and changed the code to just remove their custom stack trace from any exceptions. Would you prefer this version? It's pretty specific, but I think a lot of people are still using the better_errors gem

lib/parallel.rb
```diff
@@ -30,6 +27,11 @@ def backtrace
   class ExceptionWrapper
     attr_reader :exception
     def initialize(exception)
+      # Remove bindings stack that is added by the better_errors gem
```
grosser: `... because it is undumpable`

grosser:
prefer this ... super specific fix for a weird issue :)

grosser:
just needs a `send` to make 1.9 happy :)

ndbroadbent:
Haha just saw that, should be all green now :crossed_fingers:

lib/parallel.rb
ndbroadbent: updated the comment

grosser:
❤️ 1.11.1 :D

tjoyal:
Might have re-surfaced in some form. Still trying to pinpoint the root cause but I might jump on something else before getting to the bottom so might as well share what I have so far.

This is what I needed to remove from a  pretty fresh project.
https://github.com/tjoyal/parallel/commit/bd0f12898b92ef548ccb47545ba8bdfde622db42

```
$ rails -v
Rails 5.1.4
$ ruby -v
ruby 2.4.3p205 (2017-12-14 revision 61247) [x86_64-darwin17]
```

Hopefully this might help some until we get to understand the "recent" changes that could cause it.

PR #207, opened 2017-07-24
Fix #149 - Utilize Etc.nprocessors in Ruby 2.2+

### LINKED ISSUE #149: Why Parallel.processor_count do not use standard Etc.nprocessors?
There is a standard method in Ruby 2.2.0: [`Etc.nprocessors`](http://ruby-doc.org/stdlib-2.2.0/libdoc/etc/rdoc/Etc.html#method-c-nprocessors)

I think it would be better to rely on method from standard library instead of implemented it in system dependent way like done in [the code](https://github.com/grosser/parallel/blob/master/lib/parallel/processor_count.rb#L17]

What do you think?

Disclaimer: I have not tested all supported by `Parallel.processor_count` platforms to ensure that `Etc.nprocessors` behaves on them correctly.

westonganger:
Fixes #149

### DIFF
--- a/lib/parallel/processor_count.rb
+++ b/lib/parallel/processor_count.rb
@@ -1,3 +1,7 @@
+if RUBY_VERSION.to_f >= 2.2
+  require 'etc'
+end
+
 module Parallel
   module ProcessorCount
     # Number of processors seen by the OS and used for process scheduling.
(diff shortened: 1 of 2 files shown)


lib/parallel/processor_count.rb
```diff
@@ -16,36 +20,40 @@ module ProcessorCount
...
-          IO.popen("/usr/sbin/sysctl -n hw.ncpu").read.to_i
-        elsif File.executable?("/sbin/sysctl")
-          IO.popen("/sbin/sysctl -n hw.ncpu").read.to_i
+        if defined?(Etc) && Etc.respond_to?(:nprocessors)
+          Etc.nprocessors
         else
-          $stderr.puts "Unknown platform: " + RbConfig::CONFIG["target_os"]
-          $stderr.puts "Assuming 1 processor."
-          1
+          os_name = RbConfig::CONFIG["target_os"]
```
grosser: might be better as a new `guess_processor_count` method ...

grosser:
looks good!

grosser:
1.12.0

PR #223, opened 2018-05-23
Add ProgressBar dependency

stympy:
When using this gem via Bundler, using the `progress` option results in this error:

    LoadError: cannot load such file -- ruby-progressbar

This PR adds the dependency so that the progressbar gem will be installed along with parallel.


grosser:
I'd prefer to have fewer dependencies ... but we can make the error nicer, maybe with `gem 'ruby-progressbar'`

PR #231, opened 2018-11-18
Added note about `each_with_index` being more performant.

joshuapinter:
Compared with `start` or `finish` callbacks.

This caught me out. And the difference is significant. Something like 10x - 50x faster, depending on the record count, processes, etc.


grosser:
hmm yeah kinda obvious, but why not :)

joshuapinter:
👍 You'd think so but it caught me out, and I'm only half dumb. :)

PR #238, opened 2019-02-19
added ruby 2.6 and jruby to test matrix

### LINKED ISSUE #140: Add JRuby to the testing matrix for travis

walf443:
ruby 2.6.x was already released
https://www.ruby-lang.org/en/news/2019/01/30/ruby-2-6-1-released/


.travis.yml
```diff
@@ -12,4 +12,5 @@ rvm:
   - 2.3
   - 2.4
   - 2.5
-
+  - 2.6
```
grosser: might have to use 2.6.1 or 2.6.0 ... not sure why it does not work ... or google about supported travis ruby versions :/

grosser: ... maybe remove the dist+sudo parts ...

walf443: Thanks quick response. It may be that precise dist is too old.

walf443: I see. it should consider http://rubies.travis-ci.org/

walf443:
I'll fix this deprecation warning failure at later time https://travis-ci.org/grosser/parallel/jobs/495262000

grosser:
can either bump activesupport or set it to ignore deprecations 🤷‍♂️

walf443:
I give up to add jruby on this matrix.
It maybe better to update travis environment to Ubuntu 16.04 before this.

spec/parallel_spec.rb
```diff
@@ -188,7 +188,7 @@ def cpus
 
     it 'does not leave processes behind while running' do
       skip if ENV['TRAVIS'] # this randomly fails on travis all the time :(
-      `ruby spec/cases/closes_processes_at_runtime.rb`.should == 'OK'
+      `ruby spec/cases/closes_processes_at_runtime.rb`.split(/\n/).last.should == 'OK'
```
grosser: this is a jruby fix ... but jruby got removed ?

walf443: It's for ruby 2.6's BigDecimal warnining https://travis-ci.org/grosser/parallel/jobs/495262000

grosser:
let's do that then :)

walf443:
@grosser I'm sorry to fixed it lately. All test passed.

spec/parallel_spec.rb
grosser: would prefer to ignore that explicitly then
`.sub(/.*use BigDecimal.*\n/, "")`

grosser:
thx for trying, every little step helps :)

spec/parallel_spec.rb
walf443: fixed

spec/parallel_spec.rb
```diff
@@ -188,7 +188,7 @@ def cpus
 
     it 'does not leave processes behind while running' do
       skip if ENV['TRAVIS'] # this randomly fails on travis all the time :(
-      `ruby spec/cases/closes_processes_at_runtime.rb`.should == 'OK'
+      `ruby spec/cases/closes_processes_at_runtime.rb`.gsub(/[^\n]* deprecated; use BigDecimal[^\n]*\n/, '').should == 'OK'
```
grosser: no need for `[^\n]` since regexes don't jump lines

`"abc\nb".gsub(/a.*/, "") == "\nb"`

PR #239, opened 2019-02-23
Drop ruby support under 2.2

walf443:
https://www.ruby-lang.org/en/downloads/branches/

Ruby 2.2 was already EOL before about a year. And many libraries drop support it.

I'm trying to build development environment of this project on my machine.
But It's hard because  Gemfile.lock libraries is too old.
I'll try to update libraries before drop ruby support, but It's hard. So I try to drop old ruby version.

If this procject can drop old ruby support, I'll try to update Gemfile.lock dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ dist: precise
 branches:
   only: master
 rvm:
+  - 2.2
   - 2.3
   - 2.4
   - 2.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,6 @@ dist: precise
 branches:
   only: master
 rvm:
-  - 1.9
-  - 2.0
-  - 2.1
-  - 2.2
   - 2.3
   - 2.4
   - 2.5
--- a/lib/parallel/processor_count.rb
+++ b/lib/parallel/processor_count.rb
@@ -1,60 +1,11 @@
-if RUBY_VERSION.to_f >= 2.2
-  require 'etc'
-end
+require 'etc'
 
 module Parallel
   module ProcessorCount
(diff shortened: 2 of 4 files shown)


lib/parallel/processor_count.rb
```diff
@@ -1,60 +1,11 @@
...
-            IO.popen("/usr/sbin/sysctl -n hw.ncpu").read.to_i
-          elsif File.executable?("/sbin/sysctl")
-            IO.popen("/sbin/sysctl -n hw.ncpu").read.to_i
-          else
-            $stderr.puts "Unknown platform: " + RbConfig::CONFIG["target_os"]
-            $stderr.puts "Assuming 1 processor."
-            1
-          end
-        end
+        Etc.nprocessors
```
walf443: jruby seems to also support Etc.nproccessors since 2014. https://github.com/jruby/jruby/commit/077c4ffc52c95f34a2339768fd6bff972918fc9a#diff-0a1efa36a4878b5c382e854d8927b57c

.travis.yml
```diff
@@ -5,10 +5,6 @@ dist: precise
 branches:
   only: master
 rvm:
-  - 1.9
-  - 2.0
-  - 2.1
-  - 2.2
```
grosser: how about we keep 2.2 just to be nice ?

grosser:
this is super awesome cleanup :D
... I've carried around the 1.9 baggage for a while since lots of testing libraries need 1.9 support ... but I think these days are over and they can still install old versions, so let's try this ... can we keep 2.2 though, since dropping that does not buy us much ?

walf443:
OK. I'll keep ruby v2.2. My motivation is just for #240

PR #245, opened 2019-03-25
Calculate lap correctly to replace worker in isolation mode

aeroastro:
@grosser 

This fix the `isolation` bug in #244

Since the `lap` in the PR was a local variable, it could not correctly count the lap for each worker, and resulted in never-replaceable workers.

This patch fix the bug by using instance variable `@lap` of `Worker`.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -44,9 +44,10 @@ module Parallel
 
   class Worker
     attr_reader :pid, :read, :write
-    attr_accessor :thread
+    attr_accessor :thread, :lap
     def initialize(read, write, pid)
       @read, @write, @pid = read, write, pid
+      @lap = 0
     end
 
     def stop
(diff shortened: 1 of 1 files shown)


grosser:
ah ... tricky ... so we need a better test for isolation too ? (that fails without these changes)

aeroastro:
OK. I will add it.

aeroastro:
Actually, there are already tests on this isolation.

https://github.com/grosser/parallel/blob/a7138afd7f739a7d4e02c987e7030f86f220b521/spec/cases/map_isolation.rb#L3-L8

Current implementation also have a bug on assigning worker. I have fixed it too.

aeroastro:
Following are the simple execution of the code.

## current master

```
$ bundle exec ruby spec/cases/map_isolation.rb
1
2
1
1
OK
```

## this branch

```
$ bundle exec ruby spec/cases/map_isolation.rb
1
2
3
4
OK
```

grosser:
... I'll fix the existing testcase then

aeroastro:
Thank you for merging this ❤️

grosser:
`skip if ENV['TRAVIS'] # this randomly hangs on travis` <- oh boy :D

grosser:
thx for fixing a nasty bug :D
trying to re-enable the test here https://github.com/grosser/parallel/pull/252

grosser:
1.16.1

PR #248, opened 2019-03-26
Ensure kill child threads

Yuki-Inoue:
Currently, when parent thread is killed, the child threads/processes will remain.

I think they can be handled by simply ensuring to `kill` each thread in the `in_threads` method, since `kill` against dead thread is just an No-Op.

Any opinion?

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -205,7 +205,7 @@ module Parallel
       threads = Array.new(count) do |i|
         Thread.new { yield(i) }
       end
-      threads.map!(&:value)
+      threads.map(&:value)
     ensure
       threads.each(&:kill)
     end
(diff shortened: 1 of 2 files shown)


Yuki-Inoue:
FWIW, following code can be executed to see the behavioral change:

```
$ bundle exec ruby -e 'require "parallel"; require "timeout"; t = Thread.new { Parallel.map((0..60), in_threads: 2) { |i| p i; sleep 1 } }; sleep 1; t.kill; sleep'
```

grosser:
looks good!
can you add a test that fails without this change ?

Yuki-Inoue:
@grosser Wrote tests for threads/processes.

grosser:
1.16.0

PR #251, opened 2019-03-28
control timing of async exception

Yuki-Inoue:
In ruby, async exception can be raised at arbitrary timing.
To make sure to kill each child threads, use Thread.handle_interrupt.

Refer:
https://docs.ruby-lang.org/en/2.6.0/Thread.html#method-c-handle_interrupt

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -208,7 +208,7 @@ module Parallel
           count.times do |i|
             threads << Thread.new { yield(i) }
           end
-          Thread.handle_interrupt(Exception => :always) do
+          Thread.handle_interrupt(Exception => :immediate) do
             threads.map(&:value)
           end
         ensure
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -201,13 +201,20 @@ module Parallel
 
   class << self
     def in_threads(options={:count => 2})
-      count, _ = extract_count_from_options(options)
-      threads = Array.new(count) do |i|
-        Thread.new { yield(i) }
+      Thread.handle_interrupt(Exception => :never) do
+        begin
+          threads = []
+          count, _ = extract_count_from_options(options)
+          count.times do |i|
+            threads << Thread.new { yield(i) }
+          end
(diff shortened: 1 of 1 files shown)


grosser:
idk how this is different from the previous code ... can you add a testcase that fails without this change or explain more what kind of issue this prevents ?

Yuki-Inoue:
When `Thread#raise` method is called against a thread, (which is what Timeout.timeout does) it will result in immediate exception on the target thread, even if you are in the `ensure` clause and executing those codes inside.

So, to ensure the `ensure` clause to actually finish, even if your thread is somehow interrupted, I modified the code so that the ensure clause will not get interrupted.

Reference: https://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/

Yuki-Inoue:
For example, you can execute following code to confirm ensure clause gets killed by kill from other thread:

```
t = Thread.new do
  begin
    puts "thread started"
  ensure
    sleep 2
    puts "thread end"
  end
end
sleep 1
t.kill
puts "main end"
sleep
```

the outputs

```
thread started
main end
```

Yuki-Inoue:
With following code, you can see thread.raise cancels ensure clause

```
t = Thread.new do
  begin
    puts "thread started"
  ensure
    sleep 2
    puts "thread end"
  end
end
t.report_on_exception = false
sleep 1
t.raise
puts "main end"
sleep
## OUTPUTS
# thread started
# main end
```

whereas with following code, exception is delayed

```
t = Thread.new do
  Thread.handle_interrupt(Exception => :never) do
    begin
      Thread.handle_interrupt(Exception => :immediate) do
        puts "thread started"
      end
    ensure
      sleep 2
      puts "thread end"
    end
  end
end
t.report_on_exception = false
sleep 1
t.raise
puts "main end"
sleep
## Outputs
# thread started
# main end
# thread end
```

grosser:
thx for explaining ... still looks a bit scary to me, but seems reasonable! :)

grosser:
1.17.0

PR #253, opened 2019-03-28
prefer local variable over instance variable

grosser:
@aeroastro


aeroastro:
Thanks for the followup ❤️

grosser:
made it even better by using a boolean :D

PR #263, opened 2019-09-25
Allow to override the number of processor with an env variable

bpaquet:


grosser:
this is meant as a library, so pass it in 🤷‍♂ 
`Parallel.each(in_processes: ENV.fetch('FOO').to_i)`
... I don't really want to set this globally

bpaquet:
> this is meant as a library, so pass it in 🤷‍♂
> `Parallel.each(in_processes: ENV.fetch('FOO').to_i)`
> ... I don't really want to set this globally

I agree with you, but let me elaborate a little bit.

Few years ago, laptops  had 4. This default level of parallelization was fine. Now I have 12. My laptop starts burning everytime I start a tool using parallel  (and there are lot ). Most of them do not offer an option to tune the parallel level). It's even worse in the server world. I have servers with 96 or 144 cores. Most of the times, there is no point to parallelize at this level, the only result is memory exhaustion :( That's why I think parallel, as a recognized standard library for parallelism in ruby,  should offer a way to specify the default value of parallelism.

Let me know what you think. I will not insist more if you dont agree. I can implement another solution if it allows me to change the default value of parallelism.

grosser:
changed to PARALLEL_PROCESSOR_COUNT and added tests in https://github.com/grosser/parallel/pull/265

Issue #268, opened 2019-11-07
:process_incoming_jobs doesn't rescue non-standard error, which could lead to DeadWorker


RusHibaM:
Thanks in advance for making this Gem 😄. Our team has been enjoying it for a long time.~

However, as we also use Rspec in our automation systems, I found that sometimes our test would fail with a DeadWorker. This DeadWorker makes our investigation a little bit difficult. By digging into your library code more, I found that the fact that our tests would fail with DeadWorker is a consequence of:

1. When something like `expect{a}.to be b` fail, Rspec will raise a RSpec::Expectations::ExpectationNotMetError, which isn't a StandardError
2. We are using Parallel with `in_processes` so many of our code will be executed on a forked process.
3. The rescue in https://github.com/grosser/parallel/blob/master/lib/parallel.rb#L472 doesn't seem to catch the RSpec::Expectations::ExpectationNotMetError as it isn't a StandardError.
4. The process died out, and the worker failed to read the file. When this happened, DeadWorker from (https://github.com/grosser/parallel/blob/master/lib/parallel.rb#L74) will be the final thing recorded.

I have a small example so you could also have a try:

```ruby
require 'rspec/expectations'
require 'parallel'

# This line will fail and the last thing in the rspec report is a Parallel::DeadWorker
Parallel.each(1..1, in_processes:1) {|x| raise RSpec::Expectations::ExpectationNotMetError}

# This line will leave the correct exception as the exception happens in the main process and 'kills' the process
Parallel.each(1..1, in_threads:1) {|x| raise RSpec::Expectations::ExpectationNotMetError}

```

I tried to make a small change that lets the rescue (https://github.com/grosser/parallel/blob/master/lib/parallel.rb#L472) to catch `Exception` and after the changes, our annoying DeadWorker was resolved.

As I don't really know the designing details of process_incoming_jobs, this Issue is mainly for asking:

1. Are there any concerns so that you guys didn't choose to rescue all the exceptions in process_incoming_jobs?
2. We could patch what we need in our own system. But do you think it is worthy to modify the process_incoming_jobs so it could catch `Exception`? Or is there any other thing that might be better? We benefit from the open source community a lot so I would like to submit a PR for this. Feels like this could be something that everyone could run into, especially when using libraries such as Rspec that raises non-StandardError.

grosser:
Yeah it might be a good idea to rescue all ... since it's done in a fork and when we don't things get weird ...
[...]

PR #269, opened 2019-11-11
Make sure to also rescue non-standard error

RusHibaM:
A potential way to solve #268

I tried to raise the SystemExit again as we have this test case:

    raises DeadWorker when using exit so people learn to not kill workers and do not crash main process

If I just rescue all the exceptions, this test will fail since no DeadWorker will be raised when the worker exits.

I'm also thinking about allowing users to specify the non-standard errors that they wish to rescue. By doing this, we may not need to handle all different edge cases and just let the users choose what to rescue.

```ruby
Parallel.each([1,2,3], extra_exceptions_to_handle: [Exception]) do |index|
end
```

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -469,7 +469,7 @@ module Parallel
         item, index = job_factory.unpack(data)
         result = begin
           call_with_index(item, index, options, &block)
-        rescue SystemExit
+        rescue SystemExit, SignalException
           raise $!
         rescue Exception
           ExceptionWrapper.new($!)
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -469,7 +469,9 @@ module Parallel
         item, index = job_factory.unpack(data)
         result = begin
           call_with_index(item, index, options, &block)
-        rescue
+        rescue SystemExit
+          raise $!
+        rescue Exception
           ExceptionWrapper.new($!)
         end
         begin
--- a/spec/cases/exception_raised_in_process.rb
+++ b/spec/cases/exception_raised_in_process.rb
@@ -1,0 +1,11 @@
+require './spec/cases/helper'
+
+begin
+  Parallel.each([1]){ raise Exception }
+rescue Parallel::DeadWorker
+  puts "No, DEAD worker found"
+rescue Exception
+  puts "Yep, rescued the exception"
+else
+  puts "WHOOOPS"
+end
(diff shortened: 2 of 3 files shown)


lib/parallel.rb
```diff
@@ -469,7 +469,9 @@ def process_incoming_jobs(read, write, job_factory, options, &block)
         item, index = job_factory.unpack(data)
         result = begin
           call_with_index(item, index, options, &block)
-        rescue
+        rescue SystemExit
```
grosser:
```suggestion
        rescue SystemExit, SignalException
```
add SignalException so we do not capture the user hitting Ctrl+c ?

grosser:
we'll also need a test that explains the usecase / would not work with the old code

RusHibaM:
> we'll also need a test that explains the usecase / would not work with the old code

Sure, will add a test case. 
BTW, I noticed that `Parallel.map can work in isolation` was failing in Travis for my changes. I'm a little bit confused as it seems to be passing locally on my machine.

grosser:
1.19.0 ... let's see if this comes back to bite us :D

RusHibaM:
> 1.19.0 ... let's see if this comes back to bite us :D

Haha, let's see. 😄

valscion:
Neat change! I wonder if this list should also have `NoMemoryError` and `Interrupt` that should probably not be rescued either. This way the list of "exceptions not to rescue" would match what exceptions RSpec itself does not rescue:

https://github.com/rspec/rspec-support/blob/673133cdd13b17077b3d88ece8d7380821f8d7dc/lib/rspec/support.rb#L132-L140

```rb
    module AllExceptionsExceptOnesWeMustNotRescue
      # These exceptions are dangerous to rescue as rescuing them
      # would interfere with things we should not interfere with.
      AVOID_RESCUING = [NoMemoryError, SignalException, Interrupt, SystemExit]

      def self.===(exception)
        AVOID_RESCUING.none? { |ar| ar === exception }
      end
    end
```

Our usage of `parallel` won't really care about this, though — I merely spotted this change when updating `parallel` recently and thought you might want to know about how RSpec does a similar thing :relaxed: (PR implementing this in `rspec-support` originally is here: https://github.com/rspec/rspec-core/pull/2063 in case you're interested)

grosser:
awesome, I copied that in 1.19.1

PR #270, opened 2019-11-13
Add project metadata to the gemspec

orien:
Add `bug_tracker_uri`, `wiki_uri`, `documentation_uri`, and `source_code_uri` to the gemspec metadata.

These [project metadata](https://guides.rubygems.org/specification-reference/#metadata) will facilitate easy access to project information. The URI will be available on the [Rubygems project page](https://rubygems.org/gems/parallel), via the rubygems API, and the `gem` and `bundle` command-line tools with the next release.

### DIFF
--- a/parallel.gemspec
+++ b/parallel.gemspec
@@ -7,6 +7,12 @@ Gem::Specification.new name, Parallel::VERSION do |s|
   s.authors = ["Michael Grosser"]
   s.email = "michael@grosser.it"
   s.homepage = "https://github.com/grosser/#{name}"
+  s.metadata = {
+    "bug_tracker_uri"   => "https://github.com/grosser/#{name}/issues",
+    "documentation_uri" => "https://www.rubydoc.info/gems/#{name}/#{s.version}",
+    "source_code_uri"   => "https://github.com/grosser/#{name}/tree/v#{s.version}",
+    "wiki_uri"          => "https://github.com/grosser/#{name}/wiki",
+  }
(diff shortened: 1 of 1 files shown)


parallel.gemspec
```diff
@@ -7,6 +7,12 @@ Gem::Specification.new name, Parallel::VERSION do |s|
   s.authors = ["Michael Grosser"]
   s.email = "michael@grosser.it"
   s.homepage = "https://github.com/grosser/#{name}"
+  s.metadata = {
+    "bug_tracker_uri"   => "https://github.com/grosser/#{name}/issues",
+    "documentation_uri" => "https://www.rubydoc.info/gems/#{name}/#{s.version}",
```
grosser: there are no docs there ... prefer to leave this empty

orien: 🤔 I wonder why the readme isn't rendered on that page.

grosser: it's not rdoc or so ... but I don't want my docs there anyway, I want them close to the source so people make PRs :)

grosser:
thx!

parallel.gemspec
orien: Other gems readme.md get rendered. Take Spring for example: 
https://www.rubydoc.info/gems/spring
https://github.com/rails/spring/blob/master/README.md

I wonder if it's the capitalisation. `Readme.md` vs `README.md`.

grosser:
I'm happy it does not work, so don't fix it :D

Issue #150, opened 2015-10-22
Problem of uninitialized constant


MaximeRDY:
Hello,

Thank you for this gem. 

But I have an issue in my Rails App. My app is hosted by Heroku.
 Sometimes in my delayed job worker, I have a "NameError: uninitialized constant". 

The Stack Trace: 
```
3
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 408 in call
4
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 408 in call_with_index
5
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 288 in block (2 levels) in work_in_threads
6
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 419 in with_instrumentation
7
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 287 in block in work_in_threads
8
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 183 in block (2 levels) in in_threads
```

Do you have any idea, what's going on ? And how to fix it ? 

Thanks for your response.

grosser:
no idea ... what constant is not defined ?

the backtrace is from where parallel calls your block ... so maybe the block fails ... but unable to tell from that strange backtrace :(

MaximeRDY:
Sorry for the delay, I don't receive the notification from Github. 

It's one on my `app/services`. 

This error is really weird because it's happened randomly or maybe it's seems at the first time when the dyno worker is created. 

Because when the job failed, delayed-job restart the same one just after and I don't have this error...

Maybe it's my rails configuration :/ 

See the full stack trace:
```
NameError: uninitialized constant Transports::Flights::Koedia::RequestAvailabilityForRoutes
1
File "/app/app/services/transports/flights/find_for_return_trip.rb" line 151 in koedia_request_by_trip
2
File "/app/app/services/transports/flights/find_for_return_trip.rb" line 20 in block in flight_journeys
... 6 non-project frames
3
File "/app/vendor/bundle/ruby/2.1.0/gems/parallel-1.6.1/lib/parallel.rb" line 408 in call
4
[...]

PR #272, opened 2019-12-28
docs: Improve notes about ActiveRecord

kakra:
Fixes: https://github.com/grosser/parallel/issues/150
Signed-off-by: Kai Krakow <kai@kaishome.de>

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -70,7 +70,13 @@ Processes/Threads are workers, they grab the next piece of work when they finish
 
 ### ActiveRecord
 
-Try any of those to get working parallel AR
+Parallel can be used with ActiveRecord. However, there may be a few pitfalls
+you should know of.
+
+#### Database Connection Loss
+
+Multithreading in ActiveRecords needs connection pooling or explicit
+reconnects. Try any of those to get working parallel AR:
 
 ```Ruby
 # reproducibly fixes things (spec/cases/map_with_ar.rb)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -93,6 +99,31 @@ Parallel.each(User.all, in_processes: 8) do |user|
...
 ```
 
+You may also want to bump up your connection pool size in
+`config/database.yml`.
+
+#### Unexpected and Random "NameError: unintialized constant"
+
+There's a potential race when ActiveRecord models are going to be autoloaded
+from within Parallel code blocks. This is mostly only visible in environments
+non preloading the whole application, i.e. while migrating.
```
grosser:
```suggestion
non preloading the whole application, like development/test/migrations.
```
... FYI migrations should usually not use existing models since that will make them break when someone runs old migrations

Readme.md
```diff
@@ -93,6 +99,31 @@ Parallel.each(User.all, in_processes: 8) do |user|
...
+explicitely using `require '<modelname>'` or simply calling your class using
+`ModelName.class` just before the `Parallel.each` block.
+
+Also keep in mind that your models may require other models in turn during
+validation (or resolving references), so they may autoload more classes by
+themselves. Just add those classes, too. Take care not to miss anything
+because the problem may show up randomly and haunt you unexpectedly some
+time in the future.
+
+Reference: https://github.com/grosser/parallel/issues/150
```
grosser: please condense section this down to <1/2 the number of words

Readme.md
kakra: > ... FYI migrations should usually not use existing models since that will make them break when someone runs old migrations

Yes, that's currently one problem I'm facing. But it's actually needed because I have to migrate from a legacy database scheme to a modern one to clean up a lot of crufty code. My solution is to leave the models as compatible as I can get and put todos notes in the files to remove some stuff after deployment. Later on after deployment, we can get rid of those migrations and solely rely on `schema.rb` to rebuild an empty database.

Because migrations are huge (and it takes my dev machine around 4-5 hours to run them), I've integrated parallel to get it down to around 1 hour. I regularly need to reset the database by fetching it again from production.

Thanks for the pointer, tho. :-)

grosser:
FYI https://github.com/grosser/parallel/commit/667fa9d777321ed6a57ace0f02c30d9fea7f0e1e

kakra:
Oh, this is extremely terse... But I'm fine with it. Thanks for merging.

Issue #276, opened 2020-06-10
Net::HTTP open_timeout doesn't work with `in_threads` but does work with `in_processes`


owst:
It appears that setting open_timeout doesn't work if I use `Parallel.map` with `in_threads` vs `in_processes` with a `Net::HTTP` request. This came about from me wanting to check the health of various services in parallel (on their private IP addresses), but I'd forgotten to connect my VPN, causing the connection time outs.

The following is a standalone reproduction - my default gateway (where `10.0.0.1` is routed) seems to just drop the packets:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'parallel', '= 1.19.1'
end

require 'net/http'

def go(type, i)
  start = Time.now

  net = Net::HTTP.new("www.google.com", 80 + i)
  net.open_timeout = 1
  net.get('/')
rescue => e
  puts "#{e.class.name}: #{type} for #{net.inspect} in #{Time.now - start}"
end

Parallel.map([1], in_processes: 2) do |i|
  go("in_processes", i)
end

Parallel.map([1], in_threads: 2) do |i|
  go("in_threads", i)
end
```

When I run this, I see:
```
Net::OpenTimeout: in_processes for #<Net::HTTP 10.0.0.1:3000 open=false> in 1.007244
Errno::ETIMEDOUT: in_threads for #<Net::HTTP 10.0.0.1:3000 open=false> in 75.128787
```

What causes the difference in Exception raised, and why does the `in_threads` version timeout in ~75 seconds vs ~1 second as was set as the open-timeout?

grosser:
Yeah can confirm when using www.google.com:81
also it works correctly when doing:

```ruby
Thread.new do
  go("in_threads", 1)
end.join
```

which should be the same as the parallel call.
... also tried with `.each` to simplify further but no luck

reduced it to this:
```ruby
require 'timeout'

Parallel.each([1], in_threads: 1) do |i|
  begin
    Timeout.timeout(0.1) { sleep 0.2 }
  rescue Timeout::Error
    puts "OK"
  else
    puts "BROKEN"
  end
end
```

problem was introduced here https://github.com/grosser/parallel/pull/251

grosser:
1.19.2

PR #277, opened 2020-06-16
Allow timeout usage inside of threads

grosser:
fixes https://github.com/grosser/parallel/issues/276 /cc @owst 

introduced in https://github.com/grosser/parallel/pull/251 when all exceptions were ignored during execution even though we only wanted to ignore our own `ensure` /cc @Yuki-Inoue


grosser:
1.19.2

owst:
Great, thanks @grosser!

PR #281, opened 2020-08-02
Assign worker averagely

rueyaa332266:
Hi @grosser 

I found an issue when assigning jobs to workers.
Not happen every time, but sometime when printing worker number, few workers work obviously more than others and some workers have no job to work.

Here are some examples:
```ruby
Parallel.each(0...5, :in_processes =>5) { |i| puts "Worker: #{Parallel.worker_number}" }
Worker: 0
Worker: 0
Worker: 0
Worker: 2
Worker: 1
```
```ruby
Parallel.each(0...5, :in_processes =>5) { |i| puts "Worker: #{Parallel.worker_number}" }
Worker: 0
Worker: 1
Worker: 2
Worker: 0
Worker: 3
```

I realize that when some child process or thread works faster than others,`job_factory.next` will execute more times than others.
So the worker aren't assigned averagely. 

I fix the code by adding the max work times limit in the loop, so workers can be assigned averagely.

Let me know if you have any suggestion. Thanks!

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -350,13 +350,17 @@ module Parallel
       results = []
       results_mutex = Mutex.new # arrays are not thread-safe on jRuby
       exception = nil
+      max_work_times = Array.new(options[:count], job_factory.size/options[:count])
+      max_work_times.map!.with_index { |times,i| i <(job_factory.size % options[:count]) ? times += 1 : times }
 
       in_threads(options) do |worker_num|
         self.worker_number = worker_num
         # as long as there are more jobs, work on one of them
-        while !exception && set = job_factory.next
+        loop.with_index(1) do |_, count|
+          break if exception || (count > max_work_times[worker_num])
           begin
-            item, index = set
+            item, index = job_factory.next
+            break unless index
             result = with_instrumentation item, index, options do
               call_with_index(item, index, options, &block)
             end
@@ -375,6 +379,8 @@ module Parallel
       results = []
       results_mutex = Mutex.new # arrays are not thread-safe
       exception = nil
+      max_work_times = Array.new(options[:count], job_factory.size/options[:count])
+      max_work_times.map!.with_index { |times,i| i <(job_factory.size % options[:count]) ? times += 1 : times }
 
       UserInterruptHandler.kill_on_ctrl_c(workers.map(&:pid), options) do
         in_threads(options) do |i|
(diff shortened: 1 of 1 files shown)


grosser:
won't that make things slower if there is for example 2 workers with 1 slow item and 9 fast items ? (as in the 5 fast items will get processed and then it would wait for the slow worker to finish and do 4 more fast items)

rueyaa332266:
Totally agree with you. Worker choose the faster way to assign rather than assign averagely.
I got that, thanks!

Actually I'm working on something that I want workers to got equivalence item to work not depend on speed.
Would you mind that I will change this assign averagely feature into an option just like `isolation: true`.

grosser:
hmmm I'm not a big fan since this sounds very 1-off-usecase and adds edge-cases/maintenance/etc for other users

what exactly is the usecase that needs average items ? ... can it maybe just sleep after being done to not unblock the worker ?

rueyaa332266:
I'm making some tests with "test/unit" and creating test classes in parallel.

test/unit cases run after every worker assigned. And these cases needs some times to run. So when some workers get more cases than others, it becomes not so efficient. 

The usecase is like this:
```ruby
require 'test/unit'
require 'parallel'

class Example < Test::Unit::TestCase
end

def work(case_name)
  Example.class_eval %Q"def test_#{case_name}; p 'work'; end"
end

Parallel.each(0...4, :in_processes => 2) do |i|
    work(i)
    puts "Worker: #{Parallel.worker_number}"
end

### output ###
Worker: 1
Worker: 1
Worker: 0
Worker: 1
Loaded suite test
Started
"work"
."work"
."work"
.
Finished in 0.000694 seconds.
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
3 tests, 0 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
4322.77 tests/s, 0.00 assertions/s
Loaded suite test
Started
"work"
.
Finished in 0.00144 seconds.
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 0 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
694.44 tests/s, 0.00 assertions/s
```

rueyaa332266:
Hi, @grosser. I updated the usecase.
Do you have any advise?
Or is it not a good idea to use test/unit in parallel.

grosser:
uses https://github.com/grosser/parallel_tests or https://github.com/grosser/parallel_split_test ?

rueyaa332266:
I'll check them! Thanks for your help.
Close this pull request.

Issue #273, opened 2020-01-16
#.find


kochd:
I got a expensive calculation and i want to kill all childs once one of them delivers a result:

``` ruby
results = Parallel.find(1..32, in_processes: 32) do
  this_could_take_long(1, 2)
end
p results #=> ["solution"]
```

Use case is solving a proof of work. One of the 16 processes will be the fastest and only one solution is needed. There seams to be no way to this or is ?

grosser:
check how  `Parallel.any?` is implemented ... `.find` should be very similar ...

grosser:
... not sure if it's actually being lazy though ... there is support to `raise Parallel::Break`, just not sure how to also get the returned value ... that could be stored in the exception though, like `raise Parallel::Break, result`

ellcs:
I've made an implementation. It's a bit hacky and would be happy to get some feedback. Before i make a pull request, i'd like to mention that `Enumerable#find` does actually *not* return the result, but the found element. So we need another name. While scrolling through the [Array Documentation](https://ruby-doc.org/core-2.7.0/Array.html) and [Enumerable Documentation](https://ruby-doc.org/core-2.7.0/Enumerable.html) i did not find a matching name.

What do you think about the following names:
```ruby
block = lambda { |e| if e > 2; e; end }
Parallel.first_result([1,2,3], &block) 
Parallel.find_first_result([1,2,3], &block) 
Parallel.first_done([1,2,3], &block) 
```

grosser:
usually this is done with `[1,2,3].find { |x| break x if x.even? }` so let's implement that inferface :)

ellcs:
I don't want to start a dicussion, but... :)

I have some thoughts on your suggested interface `[1,2,3].find { |x| break x if x.even? }`:

- The `break x` does not return `x` as you can see in the example below. It still returns the found element. So you actually do not need the `break` in the block and it has the same functionallity as `{ |x| x.even? }`.
  ```ruby
  block = lambda { |x| break(x*9) if x > 5 }
  (1..10).find(&block)
  # => 6
  ```
- `find` is not intended to return the calculated value, but the element. Using that method name could lead to confusion.
- I've implemented Rubys `find` aswell, but it returns the matching element, not the result.

grosser:
`(1..10).find { break 111 }` works though 🤷

grosser:
... I guess the `break` might not be usable from inside parallel :/

grosser:
[...]

PR #283, opened 2020-10-10
Add Parallel::find and Parallel::find_result

ellcs:
Implementation of ticket #273.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -14,6 +14,14 @@ module Parallel
   class Kill < StandardError
   end
 
+  class Return < StandardError
+    attr_reader :return_value
+
+    def initialize(return_value)
+      @return_value = return_value
+    end
+  end
+
   class UndumpableException < StandardError
     attr_reader :backtrace
     def initialize(original)
@@ -228,6 +236,24 @@ module Parallel
       map(array, options.merge(:preserve_results => false), &block)
     end
 
+    def first_result(array, options = {}, &block)
+      raise "You must provide a block when calling #first_result" if block.nil?
+      map(array, options.merge({return_one: true})) do |*a| 
+        if result = block.call(*a) 
+          raise Parallel::Return.new(result)
+        end
+      end
+    end
+
+    def find(array, options = {}, &block)
+      raise "You must provide a block when calling #find" if block.nil?
+      map(array, options.merge({return_one: true})) do |*a| 
+        if block.call(*a) 
+          raise Parallel::Return.new(*a)
+        end
+      end
+    end
+
     def any?(*args, &block)
       raise "You must provide a block when calling #any?" if block.nil?
       !each(*args) { |*a| raise Parallel::Kill if block.call(*a) }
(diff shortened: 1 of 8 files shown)


lib/parallel.rb
```diff
@@ -271,14 +297,22 @@ def map(source, options = {}, &block)
       add_progress_bar!(job_factory, options)
 
       results = if size == 0
-        work_direct(job_factory, options, &block)
-      elsif method == :in_threads
-        work_in_threads(job_factory, options.merge(:count => size), &block)
-      else
-        work_in_processes(job_factory, options.merge(:count => size), &block)
-      end
+                  work_direct(job_factory, options, &block)
```
grosser, line 300: 2-space plz

lib/parallel.rb
```diff
@@ -485,7 +524,8 @@ def process_incoming_jobs(read, write, job_factory, options, &block)
     end
 
     def handle_exception(exception, results)
-      return nil if [Parallel::Break, Parallel::Kill].include? exception.class
+      return if [Parallel::Break, Parallel::Kill].include?(exception.class)
+      return exception if exception.class == Parallel::Return
```
grosser, line 528: I'd expect this to return the value ?
```suggestion
      return exception.result if exception.class == Parallel::Return
```

lib/parallel.rb
```diff
@@ -271,14 +297,22 @@ def map(source, options = {}, &block)
...
+        if options[:return_one] 
+          if results.is_a?(Return)
+            results.return_value
+          end
+        elsif options[:return_results] 
+          results
+        else
+          source
+        end
       end
```
grosser, line 316: this looks like it's doing similar kind of logic to `handle_exception` ... can it be combined ?

lib/parallel.rb
```diff
@@ -228,6 +236,24 @@ def each(array, options={}, &block)
...
+
+    def find(array, options = {}, &block)
+      raise "You must provide a block when calling #find" if block.nil?
+      map(array, options.merge({return_one: true})) do |*a| 
+        if block.call(*a) 
+          raise Parallel::Return.new(*a)
+        end
+      end
+    end
+
```
grosser, line 256: maybe let's just do the `raise Return` part first, so this is a smaller PR ?

lib/parallel.rb
```diff
@@ -14,6 +14,14 @@ class Break < StandardError
   class Kill < StandardError
   end
 
+  class Return < StandardError
+    attr_reader :return_value
```
grosser, line 18:
```suggestion
    attr_reader :value
```

grosser:
Logic looks right, but I think you stumbled into a problem area of the code there, I think we should do a cleanup first before adding a new feature, the handle_exception and results/source logic seems like they would go together

ellcs:
Thanks for the review. I will take the time to apply some changes soon.

grosser:
v1.20 has `Parallel.map([1, 2, 3]) { |i| raise Parallel::Break, i if i == 2 } == 2`

PR #284, opened 2020-10-22
Adds parallel methods example to the readme

samuelpismel:


grosser:
thx for the PR
a few issues:
- example is too long/verbose
- it's not an important feature that needs to be at the top of the readme (wiki + link in readme would be nice)
- assigning variables from inside the methods won't work in processes, so better to use `map`

PR #303, opened 2021-09-02
Avoid thousands of lsof warnings in chroot

henrich:
Just executing lsof in pbuilder/cowbuilder chroot environment causes flood of warnings. Just -l option avoids this noise.
In chroot environment, /etc/passwd is a limited one but /proc has lot of processes that are related to users in the parent environment.

-l option means "inhibits  the  conversion of user ID numbers to login names."


grosser:
thx, worked fine on mac too

PR #307, opened 2021-11-16
CI: Use matrix.task value in Rake Step

olleolleolle:
This wants to go with another change: unhook the RuboCop linting from the default Rake task.


grosser:
they are supposed to be in the default task

olleolleolle:
@grosser How about this change, then: remove the added "include" section that reruns 2.5?

olleolleolle:
Unless I'm missing something unseen here, that now does nothing extra for the matrix. 2.5 is among the tested versions.

grosser:
https://github.com/grosser/parallel/pull/308

PR #311, opened 2021-11-17
Make sure to run test cases with target Ruby version

dleidert:
In Debian we have multiple Ruby versions in parallel. The ruby binary however can only point to the default one. Thus the tests will always run with the default version. To make sure that all test cases are run with each Ruby version we currently support, we propose to rely on `RbConfig.ruby`.

### DIFF
--- a/spec/parallel_spec.rb
+++ b/spec/parallel_spec.rb
@@ -20,7 +20,7 @@ describe Parallel do
   def execute_start_and_kill(command, amount, signal = 'INT')
     t = nil
     lambda {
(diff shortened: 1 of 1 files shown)


grosser:
it's a little ugly, but not having to do the lookup via rvm/rbenv is kinda nice

grosser:
rebase and fix rubocop

dleidert:
> it's a little ugly, but not having to do the lookup via rvm/rbenv is kinda nice

Well, you could do it with a function (pass the test-file as argument) and prepend it with `RbConfig.ruby` in the function. That might also fix the line length issue reported by rubocop. Also, it means to have this centralized.

grosser:
fixed in https://github.com/grosser/parallel/pull/312

Issue #298, opened 2021-08-11
Support Ruby 3.0's Ractor / Reactor


rmondello:
Ruby 3.0's Reactors seem powerful: https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

It'd be neat, if possible, if one could use Parallel to use processes when running on Ruby 2.x and Reactors in Ruby 3.x. Could take some of the weight of parallelism using processes way down.

grosser:
They are super alpha and fragile, so PR welcome if it's simple/opt-in but not if it requires a giant rewrite.

grosser:
poked this a bit and you can't pass a block to a ractor, so best you can do is:
```ruby
Parallel.map [1,2,3], in_processes: 2 do |i|
  Ractor.new { sleep 2; receive * 2 }.send(i).take
end
```

(somehow ractors in threads don't work :/ )
... would be nice to have a wiki page for this to show some basic examples

grosser:
also tried defining fake methods with the block and calling that, but got
```
 defined with an un-shareable Proc in a different Ractor
```

... so only thing that works afaik is sending the class+name of a method and arguments

so this style could work:

```ruby
class Foo
  def self.bar(i)
    # stuff
  end
end
Parallel.map [1,2,3], in_ractors: 2, ractor: [Foo, :bar]
```

grosser:
1.22.0

PR #317, opened 2022-03-09
ractors

grosser:
- in_ractors kinda works
- need to fix map/each etc
- need to add tests (exclude from ruby <3)

see gotchas from https://github.com/grosser/parallel/issues/298

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -94,3 +94,6 @@ Security/MarshalLoad:
 
 Lint/EmptyBlock:
   Exclude: [spec/**/*.rb]
+
+Naming/MethodParameterName:
+  Exclude: [spec/**/*.rb]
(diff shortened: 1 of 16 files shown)


enumag:
Hello, is there some documentation for the ractors feature? In readme you have just this:

```
results = Parallel.map(['a','b','c'], in_ractors: 3, ractor: [SomeClass, :expensive_calculation])
```

Which is a bit confusing:
- What is that `ractor: [SomeClass, :expensive_calculation]`? Is `expensive_calculation` a method name on `SomeClass` or something else?
- Is there a way to propagate global variables into the ractor? Or is it done automatically?
- If not does the ractor at least get the `ARGV` array of the original process?
- Can I instead of `in_ractors: 3` say "use as many ractors as possible for the current hardware" (such as the total amount of CPU cores)?

grosser:
- yes
- ractors cannot use global variables
- ractors cannot use ARGV
- no, either make a PR or use `in_ractors: Etc.nprocessors` / `Parallel.physical_processor_count`

enumag:
Not very nice API in my opinion but should do the job. Thanks!

PR #333, opened 2023-03-17
Inline the methods in the ProcessorCount module

M-Yamashita01:
During code reading, we found the TODO to inline the method in the ProcessorCount module.
Therefore we inlined the methods and removed unnecessary files.


grosser:
thx!

Issue #338, opened 2023-10-24
[feature] Parallel processing with ordered output


shaicoleman:
Would it be possible to add the functionality to execute things in parallel, while printing the results in the original order?

Proof of concept example:

```ruby
require 'parallel'

ITEMS = ('A'..'Z').to_a

def perform_work(item)
  sleep(rand) * 3
end

def print_item(item)
  puts item
end

results_queue = Queue.new

# Printing thread
next_index = 0
printer_thread = Thread.new do
  while next_index < ITEMS.length
    index, result = results_queue.pop

    # If not in order, push it back for rechecking later
    next results_queue << [index, result] if index != next_index

    print_item(result)
    next_index += 1
  end
end

Parallel.each_with_index(ITEMS, in_threads: 6) do |item, index|
  perform_work(item)
  results_queue << [index, item]
end

printer_thread.join
```

grosser:
the results that `Parallel.map` returns are sorted, so you could do:
```
out = Parallel.each_with_index(ITEMS, in_threads: 6) { |item, index| "my output" }
puts out
```

capturing stdout would not work since that relies on replacing $stdout and that would not be thread local
if you want to get super hacky you could replace $stdout with a thread-aware StringIO that stores each threads output separately and then returns prints it in order when all processes have finished

shaicoleman:
That doesn't do the same thing as the example I've provided.
I'm interested in a callback to print things (or save to disk, etc.) in order as they finish executing, not at the end

grosser:
if you want to print as things are finishing you can use a finish hook [docs](https://github.com/grosser/parallel#progress--eta)

in that case it would just print the result

shaicoleman:
The finish hook will be out of order. I want a finish hook in order. Try to run the example to understand what I mean if its unclear.
`Parallel.map('A'..'Z', in_threads: 6, finish: -> (item, i, result) { puts i }) { sleep rand }`

grosser:
```
require 'parallel'

want = 0
stack = []
finish = -> (item, i, result) do
  if i == want
    puts item # print current
    want += 1
    
    # print things already ready
    stack[(i + 1)...].to_a.each do |old|
      break unless old
      puts old
      want += 1
    end
  else
    stack[i] = item # store for later
  end
end

Parallel.map('A'..'Z', in_threads: 6, finish: finish) { sleep rand }

```

shaicoleman:
Yes, that's a more elegant and efficient solution.

[...]

PR #339, opened 2023-12-11
Add :finish_in_order option

shaicoleman:
See #338 for discussion

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -641,7 +641,31 @@ module Parallel
 
     def instrument_finish(item, index, result, options)
       return unless on_finish = options[:finish]
+      return instrument_finish_in_order(item, index, result, options) if options[:finish_in_order]
       options[:mutex].synchronize { on_finish.call(item, index, result) }
+    end
+
+    def instrument_finish_in_order(item, index, result, options)
+      options[:mutex].synchronize do
+        options[:finish_items] ||= []
+        options[:finish_items_done] ||= []
+        options[:finish_index] ||= 0
+        if index == options[:finish_index]
+          # call finish for current item and any ready items
+          options[:finish].call(item, index, result)
+          options[:finish_index] += 1
+          (index + 1).upto(options[:finish_items].size).each do |old_index|
+            break unless options[:finish_items_done][old_index]
+            old_item = options[:finish_items][old_index]
+            options[:finish].call(old_item, old_index, result)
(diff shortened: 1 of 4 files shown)


lib/parallel.rb
```diff
@@ -641,9 +641,33 @@ def with_instrumentation(item, index, options)
...
       return unless on_finish = options[:finish]
+      return instrument_finish_in_order(item, index, result, options) if options[:finish_in_order]
       options[:mutex].synchronize { on_finish.call(item, index, result) }
     end
 
+    def instrument_finish_in_order(item, index, result, options)
+      options[:mutex].synchronize do
+        options[:finish_items] ||= []
+        options[:finish_items_done] ||= []
+        options[:finish_index] ||= 0
```
grosser, line 652: can these be local variables ?

lib/parallel.rb
```diff
@@ -641,9 +641,33 @@ def with_instrumentation(item, index, options)
...
+        if index == options[:finish_index]
+          # call finish for current item and any ready items
+          options[:finish].call(item, index, result)
+          options[:finish_index] += 1
```
grosser, line 656: can do this in the loop to not have the code twice ?

lib/parallel.rb
shaicoleman: You can't use local variables, as they need to maintain the state between invocations.
As they are being called via as a class method, it can't use instance variables either.

lib/parallel.rb
shaicoleman: To do it in the loop, you would have to store the item first, e.g. something like this:

```ruby
      options[:mutex].synchronize do
        options[:finish_items] ||= []
        options[:finish_items_done] ||= []
        options[:finish_index] ||= 0
        options[:finish_items][index] = item
        options[:finish_items_done][index] = true
        if index == options[:finish_index]
          index.upto(options[:finish_items].size).each do |old_index|
            break unless options[:finish_items_done][old_index]
            old_item = options[:finish_items][old_index]
            options[:finish].call(old_item, old_index, result)
            options[:finish_index] += 1
          end
        end
      end
```

grosser:
FYI refactored https://github.com/grosser/parallel/pull/340

grosser:
lmk if you can spot any more issues, otherwise I'll merge this and then release a new minor 🎉

grosser:
1.24.0

PR #340, opened 2023-12-14
polish finish_in_order

grosser:
- update docs
- fix code
- simplify code
- document code
- make test check for index + result too
- reduce memory usage
followup to https://github.com/grosser/parallel/pull/339

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -150,21 +150,20 @@ Parallel.map(1..50, progress: "Doing stuff") { sleep 1 }
 
 Use `:finish` or `:start` hook to get progress information.
  - `:start` has item and index
- - `:finish` has item, index, result
+ - `:finish` has item, index, and result
 
 They are called on the main process and protected with a mutex.
+(To just get the index, use the more performant `Parallel.each_with_index`)
 
 ```Ruby
 Parallel.map(1..100, finish: -> (item, i, result) { ... do something ... }) { sleep 1 }
 ```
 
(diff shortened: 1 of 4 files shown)


shaicoleman:
A nested array uses significantly more memory allocations, the rest of the changes look good to me
```
require 'benchmark/memory'

Benchmark.memory do |x|
  x.report('two arrays') do
   array_1 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
   array_2 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 end
  x.report('nested array') do
    array = [[1, 1], [2, 2], [3, 3], [4, 4], [5, 5], [6, 6], [7, 7], [8, 8], [9, 9], [10, 10]]
  end

  x.compare!
end
```
```
 Comparison:
          two arrays:         80 allocated
        nested array:        560 allocated - 7.00x more
```

grosser:
hehe nice! :D

I think it's not as easy to benchmark since we are doing lots of array expansions too
but agree that most likely 2 arrays are the most efficient way ... just a bit ugly, I'll take another look

grosser:
I did not find a clean way of using less arrays.
being able to do `break unless (done = options[:finish_done][i])` is kinda neat
otherwise I have to have 3 arrays, 1 for result, 1 for item, 1 for "done" which kinda sucks
and if I use a hash and check for `key?` then that's also more computation

... keeping it clean/simple until someone else has a nice idea or need for this 🤷

grosser:
1.24.0

PR #341, opened 2024-02-16
Remove unneeded ENV['TRAVIS'] from spec

MITSUBOSHI:
Since it seems that only GitHub Actions is used as CI,
I removed `ENV['TRAVIS']` which is no longer needed.


grosser (approved):
nice, thx!

PR #346, opened 2024-06-07
Drop win32ole

### LINKED ISSUE #345: Ruby 3.5 to drop `win32ole`
Ruby is continuing its path of requiring more gems to be specified. `win32ole` is next on the list (among others) and will start warning in Ruby 3.4: https://github.com/ruby/ruby/commit/f365bef0c78085b7db9b6736ae59c1657baa79a3

Usage is here, to get the number of cores on windows: https://github.com/grosser/parallel/blob/f66f10552e32681bf47f7dfd2f8183d7e0de442e/lib/parallel.rb#L331-L336

Something like the following may be a suitable replacement:
```rb
IO.popen("wmic cpu get NumberOfCores", &:read).scan(/\d+/).map(&:to_i).reduce(:+)
```

Output of that command:
`NumberOfCores  \n\n4              \n\n\n\n`

I have also openend the same report on `concurrent-ruby` at https://github.com/ruby-concurrency/concurrent-ruby/issues/1048 if you are interested. Both implementations for this are basically the same.

Earlopain:
Closes #345

I refined my initial approach for this. The PR I linked in the issue for `ruby-concurrency` got merged and released.

I tried adding CI on windows for this but didn't quite manage, the mysql gem fails to compile. I also tried adding Ruby 3.3 to the matrix but there's one memory leak test that consistently fails:
> Parallel GC does not leak memory in threads
image

So I just bump `actions/checkout` to prevent a deprecation warning about old node versions.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -328,11 +328,7 @@ module Parallel
             end
             cores.count
           when /mswin|mingw/
-            require 'win32ole'
-            result_set = WIN32OLE.connect("winmgmts://").ExecQuery(
-              "select NumberOfCores from Win32_Processor"
-            )
-            result_set.to_enum.collect(&:NumberOfCores).reduce(:+)
+            physical_processor_count_windows
           else
             processor_count
           end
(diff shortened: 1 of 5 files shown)


lib/parallel.rb
```diff
@@ -328,11 +328,23 @@ def physical_processor_count
             end
             cores.count
           when /mswin|mingw/
-            require 'win32ole'
-            result_set = WIN32OLE.connect("winmgmts://").ExecQuery(
-              "select NumberOfCores from Win32_Processor"
+            # Get-CimInstance introduced in PowerShell 3 or earlier: https://learn.microsoft.com/en-us/previous-versions/powershell/module/cimcmdlets/get-ciminstance?view=powershell-3.0
```
grosser, line 331: best extract into a new method to hide the ugliness :D

lib/parallel.rb
```diff
@@ -328,11 +328,23 @@ def physical_processor_count
...
+              'powershell -command "Get-CimInstance -ClassName Win32_Processor ' \
+              '| Select-Object -Property NumberOfCores"'
             )
-            result_set.to_enum.collect(&:NumberOfCores).reduce(:+)
+            if !result || $?.exitstatus != 0
+              # fallback to deprecated wmic for older systems
+              result = run("wmic cpu get NumberOfCores")
+            end
+            if !result || $?.exitstatus != 0
+              # Bail out if both commands returned something unexpected
```
grosser, line 341: `warn "guessing processor count"` ?

grosser (approved):
maybe ignore the mutex in the test to make it pass ?

grosser:
also good enough to just add a skip on ruby 3.3, better than having no ci at all

grosser:
lmk if you are ready to merge or want to change something still

Earlopain:
👍 I'll adress your feedback tomorrow

Earlopain:
Alright, should be good now. I bumped the sqlite3 version since this older one wouldn't compile on my other machine anymore. There's currently no released rails version that works with 2.x, so pin to 1.4 for now

grosser:
awesome work, thx!

grosser:
1.25.0

PR #347, opened 2024-06-08
Improve speed for `Get-CimInstance`

Earlopain:
Small followup for #346

In terms of the previous code, this currently does `select * from Win32_Processor` which is slower. I missed this in the other PR, this improves the speed when detection runs on windows.

The pipe is still needed since that command returns all properties anyways, they are just empty/not computed.


grosser:
nice!

grosser:
1.25.1

PR #348, opened 2024-08-08
Use cgroups aware processor count by default

y-yagi:
In containerized environments, a number of CPU cores isn't the same as the available CPUs. In this case, we need to consider cgroups.

`concurrent-ruby` now has the method to get that since v1.3.1. I think it's better to use this for setting more container environment friendly default value. Ref: https://github.com/ruby-concurrency/concurrent-ruby/pull/1038

I keep `Parallel.processor_count` as is to keep supporting setting processor count via env(`PARALLEL_PROCESSOR_COUNT`).

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -337,10 +337,10 @@ module Parallel
       end
     end
 
-    # Number of processors seen by the OS, used for process scheduling
+    # Number of processors seen by the OS or value considering CPU quota if the process is inside a cgroup,
+    # used for process scheduling
     def processor_count
-      require 'etc'
-      @processor_count ||= Integer(ENV['PARALLEL_PROCESSOR_COUNT'] || Etc.nprocessors)
+      @processor_count ||= Integer(ENV['PARALLEL_PROCESSOR_COUNT'] || available_processor_count)
     end
 
     def worker_number
@@ -694,6 +694,14 @@ module Parallel
     def instrument_start(item, index, options)
       return unless (on_start = options[:start])
       options[:mutex].synchronize { on_start.call(item, index) }
+    end
+
+    def available_processor_count
+      require 'concurrent-ruby'
+      Concurrent.available_processor_count.floor
+    rescue LoadError, NoMethodError
+      require 'etc'
+      Etc.nprocessors
     end
   end
 end
--- a/Readme.md
+++ b/Readme.md
@@ -196,6 +196,7 @@ Tips
  - [Isolation] Do not reuse previous worker processes: `isolation: true`
  - [Stop all processes with an alternate interrupt signal] `'INT'` (from `ctrl+c`) is caught by default. Catch `'TERM'` (from `kill`) with `interrupt_signal: 'TERM'`
  - [Process count via ENV] `PARALLEL_PROCESSOR_COUNT=16` will use `16` instead of the number of processors detected. This is used to reconfigure a tool using `parallel` without inserting custom logic.
+ - [Process count] `parallel` uses a number of processors seen by the OS for process count by default. If you want to use a value considering CPU quota, please add `concurrent-ruby` to your `Gemfile`.
 
 TODO
 ====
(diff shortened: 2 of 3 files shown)


lib/parallel.rb
```diff
@@ -337,10 +337,11 @@ def physical_processor_count
       end
     end
 
-    # Number of processors seen by the OS, used for process scheduling
+    # Number of processors seen by the OS or value considering CPU quota if the process is inside a cgroup,
+    # used for process scheduling
     def processor_count
-      require 'etc'
-      @processor_count ||= Integer(ENV['PARALLEL_PROCESSOR_COUNT'] || Etc.nprocessors)
+      require 'concurrent-ruby'
```
grosser, line 343: how about this: no new dependency or version conflicts with other concurrent-ruby usage
... and maybe a readme entry for "if you want x add concurrent-ruby"
```
begin
  require 'concurrent-ruby'
  Concurrent.available_processor_count.to_i
rescue LoadError
  require 'etc'
  Etc.nprocessors
end
```

grosser: maybe also use `.floor` instead of `.to_i` to make this more explicit

y-yagi: Thanks for your review.

> how about this: no new dependency or version conflicts with other concurrent-ruby 

This makes sense to me. I've updated my fix to use this approach.

lib/parallel.rb
```diff
@@ -337,10 +337,10 @@ def physical_processor_count
       end
     end
 
-    # Number of processors seen by the OS, used for process scheduling
+    # Number of processors seen by the OS or value considering CPU quota if the process is inside a cgroup,
+    # used for process scheduling
     def processor_count
-      require 'etc'
-      @processor_count ||= Integer(ENV['PARALLEL_PROCESSOR_COUNT'] || Etc.nprocessors)
+      @processor_count ||= Integer(ENV['PARALLEL_PROCESSOR_COUNT'] || available_processor_count)
```
grosser, line 343: nice side-effect that if somehow etc is broken you can set env var now :)

lib/parallel.rb
```diff
@@ -695,5 +695,13 @@ def instrument_start(item, index, options)
       return unless (on_start = options[:start])
       options[:mutex].synchronize { on_start.call(item, index) }
     end
+
+    def available_processor_count
+      require 'concurrent-ruby'
+      Concurrent.available_processor_count.floor
+    rescue LoadError, NoMethodError
```
grosser, line 702: NoMethodError is a nice touch 👍

ahorek:
this change causes "negative array size" errors because
```
> Concurrent.available_processor_count.floor
 => -1
> Concurrent.physical_processor_count
 => 6
> Concurrent.processor_count
 => 12
> Etc.nprocessors
 => 12
> Concurrent.cpu_quota.floor
 => -1
```
according to docs
cpu.cfs_quota_us = -1: Unlimited CPU time.
cpu.cfs_quota_us = 0: No CPU time (tasks are blocked).
cpu.cfs_quota_us > 0: Limited CPU time as specified.

it looks like a bug in https://github.com/ruby-concurrency/concurrent-ruby/pull/1038

grosser:
Ah great 😞
I'll revert in a bit

ahorek:
You can fall back to using Etc.nprocessors if Concurrent.available_processor_count returns an invalid value, such as -1 (negative) or 0.

grosser:
@y-yagi any idea why we'd get negative numbers ?
... and thoughts on if we should discard them to use Etc or what they mean ?

more feedback in https://github.com/grosser/parallel/issues/349
... revert, yanked 1.26.0 (and released 1.26.1 so local caches get purged too)

y-yagi:
@ahorek 
Thanks for your feedback!

@grosser 
As @ahorek described, I think this is a bug of `concurrent-ruby`. I will try to fix it first.

eregon:
`Concurrent.available_processor_count` is now fixed upstream and released: https://github.com/ruby-concurrency/concurrent-ruby/pull/1060
Sorry for the bug!

grosser:
re-released as v1.26.2 🤞

Issue #349, opened 2024-08-08
rubocop --parallel failing with "negative array size"


mark-young-atg:
Our github action to run rubocop started failing when parallel was upgraded from 1.25.1 to 1.26.0

This is the error seen
```
Run bundle exec rubocop --parallel
negative array size
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/parallel-1.26.0/lib/parallel.rb:581:in `initialize'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/parallel-1.26.0/lib/parallel.rb:581:in `new'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/parallel-1.26.0/lib/parallel.rb:581:in `create_workers'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/parallel-1.26.0/lib/parallel.rb:520:in `work_in_processes'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/parallel-1.26.0/lib/parallel.rb:291:in `map'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/parallel-1.26.0/lib/parallel.rb:235:in `each'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:103:in `warm_cache'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:76:in `run'
/home/runner/work/sm-web/sm-web/vendor/bundle/ruby/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/command/execute_runner.rb:26:in `block in execute_runner'
...snip...
```

I presume this is because of the following change that uses `Concurrent.available_processor_count.floor` to determine the number of available processors if the gem `concurrent-ruby` is installed, which in my case it is.

I can see that I can use the environment variable `PARALLEL_PROCESSOR_COUNT` to override this behaviour for my configuration but I think there is an issue here to be understood and mitigated around the behaviour in github actions. Whilst I appreciate the incorrect value may come from concurrent-ruby, I think perhaps that parallel should handle a negative value in a different way. Perhaps by failing back to using `Etc.nprocessors` as the previous version did.

[...]

PR #351, opened 2024-08-16
Ensure not to use old `concurrent-ruby`

y-yagi:
It seems that some users encounter with a negative values issue of `Concurrent.available_processor_count`.
https://github.com/grosser/parallel/issues/349#issuecomment-2290381980

To avoid the issue, this PR ensure not to use old `concurrent-ruby`. This PR also removes considering of `NoMethodError` because we can assume to use `available_processor_count` always.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -697,9 +697,10 @@ module Parallel
     end
 
     def available_processor_count
+      gem 'concurrent-ruby', '>= 1.3.4'
       require 'concurrent-ruby'
       Concurrent.available_processor_count.floor
-    rescue LoadError, NoMethodError
+    rescue LoadError
       require 'etc'
       Etc.nprocessors
     end


lib/parallel.rb
```diff
@@ -697,9 +697,10 @@ def instrument_start(item, index, options)
     end
 
     def available_processor_count
+      gem 'concurrent-ruby', '>= 1.3.4'
```
eregon, line 700: What does this raise if ran under bundle exec with an older `concurrent-ruby`?

y-yagi: `Gem::LoadError` will be raised. `Gem::LoadError` is a descendant of `LoadError`. So we can rescue it in https://github.com/grosser/parallel/blob/ab06d9e6d79a8b1375debfc348ecc19cd54e390e/lib/parallel.rb#L702

PR #356, opened 2025-04-24
Lambda in main thread and Enumerator support

### LINKED ISSUE #174: Does not work with enumerators
Hello,

This code doesn't work:

```
enumerator = File.open("/usr/share/dict/words").each_line

producer = lambda do
  begin
    enumerator.next
  rescue StopIteration
    Parallel::Stop
  end
end

Parallel.each producer, in_processes: 4 do |i|
  print "#{Process.pid} #{i}\n"
end
```

It results in `fiber called across threads (FiberError)`.

I poked around in the code; why are there 4 threads "feeding" the processes?  Why not just have the main thread pluck items from the lambda to give to an available worker?

Thanks!

### LINKED ISSUE #211: Infinite series expansion
https://github.com/grosser/parallel/blob/4fe838144d1fbf65bb3aa85c3474b538255817a9/lib/parallel.rb#L92

Parallel provides `each` and `map` methods, implying it works well with Enumerables, but it fails with the following [contrived] use case:

```ruby
require 'parallel'
require 'prime'

primes = Prime.to_enum
private_key = 12344567899

Parallel.each(primes, in_threads: 16) do |prime|
  if private_key % prime == 0
    puts "Found factor: #{prime}"
    raise Parallel::Break
  end
end
```

...because if the first argument to `Parallel.each` is not a lambda, `.to_a` is called and it tries to expand the infinite series.

### LINKED ISSUE #296: Lambda producer should be run in the main thread
Currently, if the lambda provided instead of the queue/array/enumerable it's executed in one of the threads.
As far as I can see, the only reason to provide the lambda separately from the processing block is to separate the producer to run in the main thread.
Example (from real life, but simplified): 
suppose `user.generate_update` takes 0.1s and MUST happen in the main thread due to DB connections limit
But `send_update_api(params)` takes 1s and so I want to parallelize the net calls in this case, so my idea for the code would be:

```
Parallel.each( -> { user = users.shift; user&.generate_update || Parallel::Stop }, in_threads: 10) do |params|
  send_update_api(params)
end
```

But the `generate_update` also runs in the threads and so I need to use another implementation, such as [FixedThreadPool](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/docs-source/thread_pools.md), or write my own solution.

takahiro-blab:
- Change the lambda to be invoked on the main thread (closes #296)
- Add support for an infinite `Enumerator` (without calling `.to_a` method) (closes #174, closes #211)
- Change the lambda to be able to stop operations not only returning `Parallel::Stop`, but also raising `StopIteration`.

Code summary:
1. If `JobFactory` has a lambda, it now has one queue (`@runloop_queue`)
2. Worker threads push their own queue to `JobFactory`'s queue.
3. `JobFactory` pushes jobs to threads' queue in main (caller of `Parallel` singleton method) thread.
4. `Parallel.in_threads` method counts finished threads. After all worker threads have finished, last worker thread will call `JobFactory#stopper`. It stops the loop in caller thread.
5. If worker threads don't have their queue, (Ex. `Parallel.work_in_ractors`) `JobFactory` will operate as before.
6. If the source can be accessed by the index (like `Array`), `JobFactory` will operate as before.

Thanks to the clean design of the original code, these were possible with minimal changes.

Existing RSpec tests have been all passed in my environment.
(I have run test in CRuby 3.4.3 on x86_64 machine except pending tests.)

Limitations:
- These change have not been benchmarked, so it's possible that it may introduce some performance degradation.
- The code has only been tested with CRuby (MRI), and has not been verified on other Ruby implementations such as JRuby.

This implementation uses several `Thread::Queue` instances. While this may not be the most optimal approach, it helps address a few specific issues.

I have added some test codes, but I wasn't completely sure where the test should be placed in `parallel_spec.rb`, so I added it where it seemed to fit best. Feel free to move or adjust it if you think there's a better spot.

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -97,20 +97,26 @@ module Parallel
 
   class JobFactory
     def initialize(source, mutex)
-      @lambda = (source.respond_to?(:call) && source) || queue_wrapper(source)
-      @source = source.to_a unless @lambda # turn Range and other Enumerable-s into an Array
+      @lambda = enum_wrapper(source) || (source.respond_to?(:call) && source) || queue_wrapper(source)
+      @source = source.to_a unless @lambda # turn non-Enumerable-s into an Array
+      @runloop_queue = Thread::Queue.new if @lambda
       @mutex = mutex
       @index = -1
       @stopped = false
     end
 
-    def next
-      if producer?
+    def next(queue_for_thread = nil)
+      if @runloop_queue && queue_for_thread
+        return if @stopped
+        item = runloop_enq(queue_for_thread)
+        return if item == Stop
+        index = @index += 1
+      elsif producer?
         # - index and item stay in sync
         # - do not call lambda after it has returned Stop
         item, index = @mutex.synchronize do
           return if @stopped
-          item = @lambda.call
+          item = call_lambda
           @stopped = (item == Stop)
           return if @stopped
           [item, @index += 1]
@@ -121,6 +127,25 @@ module Parallel
         item = @source[index]
       end
       [item, index]
+    end
+
+    def runloop
+      return unless @runloop_queue
+
+      loop do
+        queue = @runloop_queue.pop
+        return if queue == Stop
+        item = call_lambda
+        queue.push(item)
+        break if item == Stop
+      end
+      @stopped = true
+      begin
+        while queue = @runloop_queue.pop(true)
+          queue.push(Stop) if queue != Stop # Unlock waiting threads.
+        end
+      rescue ThreadError # All threads are unlocked.
+      end
     end
 
     def size
@@ -142,7 +167,20 @@ module Parallel
       producer? ? data : [@source[data], data]
     end
 
+    def stopper = @runloop_queue&.push(Stop)
+
     private
+
+    def call_lambda
+      @lambda.call
+    rescue StopIteration
+      Stop
+    end
+
+    def runloop_enq(queue_for_thread)
+      @runloop_queue.push(queue_for_thread)
+      queue_for_thread.pop # Wait until @lambda returns.
+    end
 
     def producer?
       @lambda
@@ -150,6 +188,11 @@ module Parallel
 
     def queue_wrapper(array)
       array.respond_to?(:num_waiting) && array.respond_to?(:pop) && -> { array.pop(false) }
+    end
+
+    def enum_wrapper(source)
+      # Convert what is inaccessible by the index
+      !source.respond_to?(:[]) && source.respond_to?(:next) && source.method(:next)
     end
   end
 
@@ -211,13 +254,24 @@ module Parallel
   class << self
     def in_threads(options = { count: 2 })
       threads = []
-      count, = extract_count_from_options(options)
+      count, options = extract_count_from_options(options)
+      finished_monitor = options[:runloop] && Queue.new(1..(count - 1)) # Insert values, one less in count than the number of threads.
+      stopper = options[:stopper]
 
       Thread.handle_interrupt(Exception => :never) do
         Thread.handle_interrupt(Exception => :immediate) do
           count.times do |i|
-            threads << Thread.new { yield(i) }
+            threads << Thread.new do
+              yield(i)
+            ensure
+              begin
+                finished_monitor&.pop(true) # This must be executed even if the worker thread is killed (by #work_in_processes).
+              rescue ThreadError # Queue#pop raises ThreadError when the queue is empty.
+                stopper&.call # Stop JobFactory#runloop
+              end
+            end
           end
+          options[:runloop]&.call # Invoke lambda in caller thread, and provide jobs to thread queue.
           threads.map(&:value)
         end
       ensure
@@ -431,10 +485,11 @@ module Parallel
       results_mutex = Mutex.new # arrays are not thread-safe on jRuby
       exception = nil
 
-      in_threads(options) do |worker_num|
+      in_threads(options.merge(runloop: job_factory.method(:runloop), stopper: job_factory.method(:stopper))) do |worker_num|
+        queue_for_thread = Thread::Queue.new
         self.worker_number = worker_num
         # as long as there are more jobs, work on one of them
-        while !exception && (set = job_factory.next)
+        while !exception && (set = job_factory.next(queue_for_thread))
           begin
             item, index = set
             result = with_instrumentation item, index, options do
@@ -523,15 +578,16 @@ module Parallel
       exception = nil
 
       UserInterruptHandler.kill_on_ctrl_c(workers.map(&:pid), options) do
-        in_threads(options) do |i|
+        in_threads(options.merge(runloop: job_factory.method(:runloop), stopper: job_factory.method(:stopper))) do |i|
           worker = workers[i]
           worker.thread = Thread.current
+          queue_for_thread = Thread::Queue.new
           worked = false
 
           begin
             loop do
               break if exception
-              item, index = job_factory.next
+              item, index = job_factory.next(queue_for_thread)
               break unless index
 
               if options[:isolation]
--- a/spec/cases/infinite_sequence.rb
+++ b/spec/cases/infinite_sequence.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# Reproduction case based on GitHub Issue #211
+# Original code provided by @cyclotron3k in the issue
+
+require 'prime'
+require './spec/cases/helper'
+
+private_key = 12344567899
+
+results = []
+
+[{ in_threads: 2 }, { in_threads: 0 }].each do |options|
+  primes = Prime.to_enum
+  Parallel.each(primes, options) do |prime|
+    if private_key % prime == 0
+      results << prime.to_s
+      raise Parallel::Break
+    end
+  end
+end
+
+print results.join(',')
--- a/spec/cases/lambda_call_same_thread.rb
+++ b/spec/cases/lambda_call_same_thread.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+require './spec/cases/helper'
+
+runner_thread = nil
+all = [3, 2, 1]
+my_proc = proc {
+  runner_thread ||= Thread.current
+  if Thread.current != runner_thread
+    raise "proc is called in different thread!"
+  end
+
+  all.pop || Parallel::Stop
+}
+
+class Callback
+  def self.call(x)
+    $stdout.sync = true
+    "ITEM-#{x}"
+  end
+end
+puts(Parallel.map(my_proc, in_threads: 2) { |(i, _id)| Callback.call i })
--- a/spec/cases/lambda_can_stop_by_exception.rb
+++ b/spec/cases/lambda_can_stop_by_exception.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+require './spec/cases/helper'
+
+def generate_proc
+  count = 0
+  proc {
+    raise StopIteration if 3 <= count
+    count += 1
+  }
+end
+
+class Callback
+  def self.call(x)
+    $stdout.sync = true
+    "ITEM-#{x}"
+  end
+end
+
+[{ in_processes: 2 }, { in_threads: 2 }, { in_threads: 0 }].each do |options|
+  puts(Parallel.map(generate_proc, options) { |(i, _id)| Callback.call i })
+end
--- a/spec/parallel_spec.rb
+++ b/spec/parallel_spec.rb
@@ -722,6 +722,10 @@ describe Parallel do
     end
   end
 
+  it "can process infinite sequence enumerator" do
+    ruby("spec/cases/infinite_sequence.rb").split(',').should == ['139'] * 2
+  end
+
   it "fails when running with a prefilled queue without stop since there are no threads to fill it" do
     error = (RUBY_VERSION >= "2.0.0" ? "No live threads left. Deadlock?" : "deadlock detected (fatal)")
     ruby("spec/cases/fatal_queue.rb 2>&1").should include error
--- a/spec/parallel_spec.rb
+++ b/spec/parallel_spec.rb
@@ -730,6 +730,10 @@ describe Parallel do
     ruby("spec/cases/lambda_can_stop_by_exception.rb").should == "ITEM-1\nITEM-2\nITEM-3\n" * 3
   end
 
+  it "must call lambda in same thread" do
+    ruby("spec/cases/lambda_call_same_thread.rb").should == "ITEM-1\nITEM-2\nITEM-3\n"
+  end
+
   it "fails when running with a prefilled queue without stop since there are no threads to fill it" do
     error = (RUBY_VERSION >= "2.0.0" ? "No live threads left. Deadlock?" : "deadlock detected (fatal)")
     ruby("spec/cases/fatal_queue.rb 2>&1").should include error
--- a/spec/parallel_spec.rb
+++ b/spec/parallel_spec.rb
@@ -726,6 +726,10 @@ describe Parallel do
     ruby("spec/cases/infinite_sequence.rb").split(',').should == ['139'] * 2
   end
 
+  it "can be finished by lambda raising StopIteration" do
+    ruby("spec/cases/lambda_can_stop_by_exception.rb").should == "ITEM-1\nITEM-2\nITEM-3\n" * 3
+  end
+
   it "fails when running with a prefilled queue without stop since there are no threads to fill it" do
     error = (RUBY_VERSION >= "2.0.0" ? "No live threads left. Deadlock?" : "deadlock detected (fatal)")
     ruby("spec/cases/fatal_queue.rb 2>&1").should include error


lib/parallel.rb
```diff
@@ -97,20 +97,26 @@ def wait
...
       @stopped = false
     end
 
-    def next
-      if producer?
+    def next(queue_for_thread = nil)
+      if @runloop_queue && queue_for_thread
+        return if @stopped
+        item = runloop_enq(queue_for_thread)
+        return if item == Stop
```
grosser, line 112: why is it not like this ?
```suggestion
        item = runloop_enq(queue_for_thread)
        @stopped = (item == Stop)
        return if @stopped
```

lib/parallel.rb
```diff
@@ -123,6 +129,25 @@ def next
       [item, index]
     end
 
+    def runloop
```
grosser, line 132: would this make sense ?
```suggestion
    def run_runloop
```

lib/parallel.rb
```diff
@@ -123,6 +129,25 @@ def next
...
+
+      loop do
+        queue = @runloop_queue.pop
+        return if queue == Stop
+        item = call_lambda
+        queue.push(item)
+        break if item == Stop
+      end
+      @stopped = true
+      begin
```
grosser, line 143: is that correct ?
```suggestion
      # clear out all work queues by adding a "stop" to them which will stop the thread working on them
      begin
```

lib/parallel.rb
```diff
@@ -142,15 +167,33 @@ def unpack(data)
       producer? ? data : [@source[data], data]
     end
 
+    def stopper = @runloop_queue&.push(Stop)
```
grosser, line 170: just to keep style consistent:
```suggestion
    def stopper
      @runloop_queue&.push(Stop)
    end
```

grosser: also should this not be called `stop` ?

lib/parallel.rb
```diff
@@ -142,15 +167,33 @@ def unpack(data)
...
+
     def producer?
       @lambda
     end
 
     def queue_wrapper(array)
       array.respond_to?(:num_waiting) && array.respond_to?(:pop) && -> { array.pop(false) }
     end
+
+    def enum_wrapper(source)
```
grosser, line 193: maybe give some examples of what types this is trying to detect

lib/parallel.rb
```diff
@@ -211,13 +254,24 @@ def restore_interrupt(old, signal)
...
 
       Thread.handle_interrupt(Exception => :never) do
         Thread.handle_interrupt(Exception => :immediate) do
           count.times do |i|
-            threads << Thread.new { yield(i) }
+            threads << Thread.new do
+              yield(i)
+            ensure
+              begin
+                finished_monitor&.pop(true) # This must be executed even if the worker thread is killed (by #work_in_processes).
```
grosser, line 268: explain why it needs to be executed

lib/parallel.rb
```diff
@@ -431,10 +485,11 @@ def work_in_threads(job_factory, options, &block)
       results_mutex = Mutex.new # arrays are not thread-safe on jRuby
       exception = nil
 
-      in_threads(options) do |worker_num|
+      in_threads(options.merge(runloop: job_factory.method(:runloop), stopper: job_factory.method(:stopper))) do |worker_num|
```
grosser, line 488:
```suggestion
      thread_options = options.merge(runloop: job_factory.method(:runloop), stopper: job_factory.method(:stopper))
      in_threads(thread_options) do |worker_num|
```

lib/parallel.rb
```diff
@@ -523,15 +578,16 @@ def work_in_processes(job_factory, options, &blk)
       exception = nil
 
       UserInterruptHandler.kill_on_ctrl_c(workers.map(&:pid), options) do
-        in_threads(options) do |i|
+        in_threads(options.merge(runloop: job_factory.method(:runloop), stopper: job_factory.method(:stopper))) do |i|
```
grosser, line 581: can we reuse the options from line 408 ?

spec/cases/infinite_sequence.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# Reproduction case based on GitHub Issue #211
+# Original code provided by @cyclotron3k in the issue
```
grosser, line 4:
```suggestion
# Original code provided by @cyclotron3k in the issue
# using a enum that is infinite, so this will hang forever when trying to convert to an array
```

grosser:
thanks, looks all very well thought out to fix these edge-cases :)

mostly looks good, but more comments would help make this easier to understand
`runloop` is kinda vague, if you have a better name that would be great

lib/parallel.rb
takahiro-blab: `@stopped` will be set in `JobFactory#runloop` .

This `JobFactory#next` may be called from some method at the same time.
So your `@stopped = (item == Stop)` needs exclusive control `Mutex#synchronize` .

Taking the value out of the `@lambda` and Setting the result to`@stopped` must be handled in the critical section or  be handled in one thread, I think.
Otherwise, the certain thread may clear `@stopped` flag. This could be a bug.

So previous implementation of Parallel uses `@mutex.synchronize` .
This PR's code handles `@lambda` and check `item == Stop` in `#runloop` by one thread, so `@mutex` is given, but it's not used.

lib/parallel.rb
takahiro-blab: You're right. Please feel free to change anything.

lib/parallel.rb
takahiro-blab: Yes. By sending `Parallel::Stop` to worker threads which are waiting in `JobFactory`'s queue, worker threads will finish. When `JobFactory`'s queue is empty, a `ThreadError` will occur. Then, `JobFactory#runloop` will finish.

lib/parallel.rb
takahiro-blab: I just used the one-line syntax for brevity, but no strong preference — happy to go with the multi-line version for consistency.

And, you're right. This `#stopper` method can be renamed to` #stop` .

lib/parallel.rb
takahiro-blab: It is easier to understand that way.

Well, I thought about not using `Method` objects, but passing `JobFactory` instance to `#in_threads`'s argument...
`#in_threads` don't need the `JobFactory` instance itself, it needs some methods of `JobFactory`, so I had written codes using `Method` object like that...

lib/parallel.rb
takahiro-blab: Sorry, where is line 408?
If you mean "`thread_options = options.merge(runloop: job_factory.method(:runloop), stopper: job_factory.method(:stopper))`" of above discussion, of course we can cut out this part to a variable and reuse it.

lib/parallel.rb
```diff
@@ -523,15 +578,16 @@ def work_in_processes(job_factory, options, &blk)
...
           worker = workers[i]
           worker.thread = Thread.current
+          queue_for_thread = Thread::Queue.new
           worked = false
 
           begin
             loop do
               break if exception
-              item, index = job_factory.next
+              item, index = job_factory.next(queue_for_thread)
```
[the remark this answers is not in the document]
takahiro-blab: Value of `Thread#[]` will vary by `Fiber` change, so it may be better to use `Thread#thread_variable_get` and `#thread_variable_set`. Whether or not, using thread local variables may take care of the queue more simply.

spec/cases/infinite_sequence.rb
takahiro-blab: This test makes sure that this Pull Request's `Parallel` can handle infinite `Enumerator`.
In other words, this test never ends in previous implementation, but this PR makes `Parallel` being able to complete this test...

So if a comment were added,  how about “Parallel can handle infinite enumerator, without converting to an array”, etc.

> `# using a enum that is infinite, so this will hang forever when trying to convert to an array`

I have wanted to make sure "will [not] hang" with this test ...

takahiro-blab:
Thank you for your comments and reviews.
Handling threads is very difficult.

Certainly, `runloop` may be vague.
How about `poploop`, `lambda_loop`, `feeder_loop`?

lib/parallel.rb
takahiro-blab: Oh, `Parallel` supports Ruby 2.7 yet, doesn't it? Ruby 2.7 doesn't support one-line method definition.
Sorry, then please fix it.

lib/parallel.rb
takahiro-blab: I have fixed and pushed this change.

lib/parallel.rb
takahiro-blab: I have fixed and pushed the change in the same way as the other calling. Is this correct?

lib/parallel.rb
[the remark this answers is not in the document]
takahiro-blab: I have tried making the change which use `Thread#thread_variable_get` and `#thread_variable_set` for the queue handling, and I have pushed it.

takahiro-blab:
I have created a commit accepting some of the suggestions.
In addition, I have committed changes to `Readme.md` related to this pull request.

lib/parallel.rb
grosser: ah thx, yeah this is a tricky section :)
can you leave a bit of inline comment for the gotchas

lib/parallel.rb
```diff
@@ -123,6 +130,26 @@ def next
       [item, index]
     end
 
+    def runloop
```
grosser, line 133:
```suggestion
    def consume_enumerator_queues
```
does this work ?

```
# consume items for from enumerator queues until they stop producing
```

lib/parallel.rb
```diff
@@ -142,15 +169,35 @@ def unpack(data)
...
+
     def producer?
       @lambda
     end
 
     def queue_wrapper(array)
       array.respond_to?(:num_waiting) && array.respond_to?(:pop) && -> { array.pop(false) }
     end
+
+    def enum_wrapper(source)
```
grosser, line 197:
```suggestion
    # returns the `next` method if the source in an enumerator
    def enum_wrapper(source)
```

lib/parallel.rb
grosser: thx, can you leave a bit of this inline for future archeologists :)

grosser:
would producer_queues or enumerator_queues make sense ?
and then `consume_from_<x>_queues` as method

takahiro-blab:
Sorry I am not very good at English,
> would producer_queues or enumerator_queues make sense ?

but is this a reference to `JobFactory`'s instance variable `@runloop_queue` ?

If so, `@runloop_queue` is a line where workers stand in to get a job. (Actually workers push their own queue to `@runloop_queue` )

So do these make sense?
Instance variable `@runloop_queue`  ->  `@workers_line_queue`
Method `runloop`  -> `distribute_work`

grosser:
maybe `worker_queues` ?

`distribute_work` -> `consume_worker_queue` ?

if the method will be used by everything then something general like `distribute_work` is fine,
but if it only deals with the enumerator queues, then I'd like to be specific and consistent (use a consistent prefix/suffix like enumerator_queues) so make it clear which part of the codebase belongs to that feature

lib/parallel.rb
takahiro-blab: I noticed the comment says "`enumerator queues`", but as far as I can tell, there's only one queue being consumed from in this method.
Does this `enumerator queues` mean `JobFactory`'s `@lambda`, doesn't it?
Should this be singular instead?

lib/parallel.rb
takahiro-blab: I have added comment to this code.

lib/parallel.rb
```diff
@@ -211,13 +254,24 @@ def restore_interrupt(old, signal)
   class << self
     def in_threads(options = { count: 2 })
       threads = []
-      count, = extract_count_from_options(options)
+      count, options = extract_count_from_options(options)
+      finished_monitor = options[:runloop] && Queue.new(1..(count - 1)) # Insert values, one less in count than the number of threads.
```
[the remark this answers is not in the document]
takahiro-blab: I have added comment to this code with below question.

lib/parallel.rb
takahiro-blab: I have added comment with above question.

takahiro-blab:
I have renamed JobFactory's instance variable `@runloop_queue` and `#runloop` method.
And I have fixed where CI of Ruby 2.7 was failing.

grosser:
thx for all the update ✨ 

I can make a separate PR to get rid of ruby 2.7 if that is causing issues, meant to do that for a while but never really broke anything.

Can you rename the last few runloop leftovers to worker_queue or something similar ?

Can it check `queue.empty?` instead of relying on an exception ? (because they are expensive and can lead to warnings)

takahiro-blab:
> Can it check queue.empty? instead of relying on an exception ? (because they are expensive and can lead to warnings)

Well, without exception, a mutex is necessary. Like this:

```ruby
# In Parallel.in_threads method
if options[:runloop]
  finished_monitor = Queue.new # In Ruby 3.0 or earlier, Queue#initialize doesn't receive initial values.
    (1..(count - 1)).each { |i| finished_monitor.push(i) }
  end
  runloop_stopper = options[:stopper]
  mutex = Mutex.new # ADD MUTEX

  Thread.handle_interrupt(Exception => :never) do
    Thread.handle_interrupt(Exception => :immediate) do
      count.times do |i|
        threads << Thread.new do
          yield(i)
        ensure
          mutex.synchronize { # Add critical section
            if finished_monitor
              if finished_monitor.empty?
                runloop_stopper&.call
              else
                finished_monitor.pop
              end
            end
          }
        end
##### Omitted below #####
```

It's bacause some threads would check whether the queue is empty and call `Queue#pop` at the same time.
Without a mutex, after a thread check whether the queue is empty, other thread may check and call `Queue#pop`. It will cause race condition. So a mutex is necessary. And the mutex overhead must be considered.
`Queue#pop(true)` is probably atomic in Ruby level, so additional mutex isn't necessary.

However, thinking about it, if a mutex is used, it's not necessary to use a `Queue`, which is thread-safe. A mutex and a simple counter variable are enough.
So, the above code can be rewritten like this:

```ruby
# In Parallel.in_threads method
if options[:runloop]
  counter = count - 1 # COUNTER
  runloop_stopper = options[:stopper]
  mutex = Mutex.new # ADD MUTEX

  Thread.handle_interrupt(Exception => :never) do
    Thread.handle_interrupt(Exception => :immediate) do
      count.times do |i|
        threads << Thread.new do
          yield(i)
        ensure
          mutex.synchronize { # Add critical section
            if counter <= 0
                runloop_stopper&.call
            end
            counter -= 1
          }
        end
##### Omitted below #####
```

I haven't tested these two above. However I think my original implementation, which uses one queue and exception, is simpler code. There is no one answer because this is a subjective issue, isn't it?
Which do you think is better?

You are welcome to push any code you think is good, not just this one, also about the names of variables, to the topic branch.

takahiro-blab:
After considering, I think using a mutex and a counter is good, so I have pushed the change.

lib/parallel.rb
```diff
@@ -123,6 +133,26 @@ def next
       [item, index]
     end
 
+    def consume_worker_queue
```
grosser, line 136:
```suggestion
    def consume_worker_queues
```

lib/parallel.rb
```diff
@@ -142,15 +172,36 @@ def unpack(data)
...
+
     private
 
+    def call_lambda
+      @lambda.call
+    rescue StopIteration
+      Stop
+    end
+
+    def worker_queues_enq(queue_for_thread)
```
grosser, line 187: would this also work if we did:
```
def worker_queues_enq
  queue = Thread::Queue.new
  @worker_queues.push(queue)
  queue.pop # Wait until @lambda to give us an item to work on
end
```
so the caller has less state to take care of

grosser: ... also maybe method name `wait_for_item` ?

lib/parallel.rb
```diff
@@ -211,13 +262,28 @@ def restore_interrupt(old, signal)
...
       Thread.handle_interrupt(Exception => :never) do
         Thread.handle_interrupt(Exception => :immediate) do
           count.times do |i|
-            threads << Thread.new { yield(i) }
+            threads << Thread.new do
+              yield(i)
+            ensure
+              mutex&.synchronize do
+                if counter <= 1
+                  # Last one thread calls consume_worker_queue_stopper
```
grosser, line 279:
```suggestion
                  # last thread needs to stop the worker queue processing
```

lib/parallel.rb
takahiro-blab: Your code also works. However, well, your code makes as many queues as jobs. Those queues remain in memory until they are collected by the GC. I do not consider creating a new queue for every jobs a good implementation.
Furthermore, when we look back on it later, we will wonder “Why does this code create a new queue here?”, wouldn't we think?

In a frank implementation, I would think the queues would be reused, but why make a queue every time?

grosser:
rubocop needs a small fix, otherwise looks good 🤞

Issue #369, opened 2026-04-13
Replaced worker processes are not killed when interrupted


wintonpc:
With `in_processes` > 0 and `isolation: true`, when the interrupt signal is received, the replacement worker processes are waited but not killed. Only the original set of worker processes are sent SIGKILL. Shutdown hangs until the replacement workers have completed their pending items normally.

The bug is masked in typical usage. Pressing Ctrl+C in a terminal sends SIGINT to the foreground _process group_, which means the replacement workers receive SIGINT and terminate relatively quickly. By contrast, `kill -INT <pid>` sends SIGINT to only the parent process, and shutdown hangs until the replacement workers exit normally.

[This PR](https://github.com/grosser/parallel/pull/370) has a test to reproduce the bug and a possible fix.

My particular use case is that I want the first Ctrl+C to initiate graceful shutdown and a second Ctrl+C to terminate immediately. The current solution is to `trap(:INT)` in my code and set `interrupt_signal: "TERM"`. The first time the SIGINT handler runs, it sets a flag to make the `start` hooks raise `Parallel::Break`. The second time, it calls `Process.kill("TERM", Process.pid)` to instruct `Parallel` to start an orderly shutdown, which is where the hang happens. The workaround is easy enough: call `Process.kill("TERM", 0)` instead so any workers that don't receive SIGKILL will at least get SIGTERM.

**System details**
- parallel 2.0.1
- Ruby 4.0.1
- GNU bash 5.2.37
- Ubuntu 25.10

PR #370, opened 2026-04-13
Kill replaced worker processes when interrupted

wintonpc:
Related issue: https://github.com/grosser/parallel/issues/369

I preserved the existing contract as much as I could, but there's a non-zero risk of breaking code that uses the gem. Given the niche use case and easy work around, I'm not sure the reward outweighs the risk. (Although since parallel just bumped to 2.x, it may be close to an appropriate time for a potentially breaking change.)

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -168,45 +168,20 @@ module Parallel
     INTERRUPT_SIGNAL = :SIGINT
 
     class << self
-      # kill all these pids or threads if user presses Ctrl+c
+      # kill all these workers if user presses Ctrl+c
       def kill_on_ctrl_c(workers, options)
         @to_be_killed ||= []
-        old_interrupt = nil
+        old_handler = nil
         signal = options.fetch(:interrupt_signal, INTERRUPT_SIGNAL)
 
-        if @to_be_killed.empty?
-          # Wrap the existing interrupt handler to kill the workers first. Workers may have been replaced,
-          # so get the latest pids.
-          # 1. The worker arrays in @to_be_killed are protected by options[:mutex].
-          # 2. Mutexes cannot be obtained in a trap context.
-          # 3. To preserve semantics, the workers must be killed before the old handler runs.
-          # 4. To preserve semantics, the old handler must run in a trap context on the main thread.
-          kill_thread = nil
-          old_interrupt = Signal.trap(signal) do
-            next if kill_thread
-            warn 'Parallel execution interrupted, exiting ...'
-            kill_thread = Thread.new do
-              pids = options[:mutex].synchronize do
-                @to_be_killed.flatten(1).map(&:pid)
-                # FUTURE: stop JobFactory from spawning new workers
-              end
-              pids.each { |pid| kill(pid) }
-              if old_interrupt == "DEFAULT"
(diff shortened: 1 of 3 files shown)


lib/parallel.rb
```diff
@@ -169,24 +169,44 @@ class UserInterruptHandler
...
+          kill_thread = nil
+          old_interrupt = Signal.trap(signal) do
+            next if kill_thread
             warn 'Parallel execution interrupted, exiting ...'
-            @to_be_killed.flatten.each { |pid| kill(pid) }
-          end
+            kill_thread = Thread.new do
+              pids = options[:mutex].synchronize do
+                @to_be_killed.flatten(1).map(&:pid)
+                # FUTURE: stop JobFactory from spawning new workers
```
wintonpc: Doing this properly would involve a new `Mutex#synchronize` around both the "stopped" check and the call to `replace_worker`, which I'm hesitant to add without understanding the code better.

spec/parallel_spec.rb
```diff
@@ -144,7 +144,13 @@ def cpus
       end.should <= 4
     end
 
-    it "preserves original intrrupts" do
+    it "kills replaced workers when handling the interrupt signal" do
+      time_taken do
+        ruby("spec/cases/isolated_interrupt.rb 2>&1")
+      end.should <= 2
```
grosser, line 151: an interrupted process still finishing successfully is a bit sus

spec/cases/isolated_interrupt.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
grosser, line 1: some description here what this is doing/simulating

lib/parallel.rb
```diff
@@ -169,24 +169,44 @@ class UserInterruptHandler
...
       # kill all these pids or threads if user presses Ctrl+c
-      def kill_on_ctrl_c(pids, options)
+      def kill_on_ctrl_c(workers, options)
         @to_be_killed ||= []
         old_interrupt = nil
         signal = options.fetch(:interrupt_signal, INTERRUPT_SIGNAL)
 
         if @to_be_killed.empty?
-          old_interrupt = trap_interrupt(signal) do
+          # Wrap the existing interrupt handler to kill the workers first. Workers may have been replaced,
```
grosser: this mess should go into a nicely explained method

grosser:
no super excited for this ... killed process just continuing could also be seen as a bug

spec/parallel_spec.rb
wintonpc: It doesn't finish successfully. I've added an assertion to that effect.

spec/cases/isolated_interrupt.rb
wintonpc: Added.

lib/parallel.rb
wintonpc: Good call. Done.

wintonpc:
> no super excited for this ... killed process just continuing could also be seen as a bug

No problem. As I mentioned, there's a reasonable workaround. I just wanted to submit a best effort fix along with the [bug report](https://github.com/grosser/parallel/issues/369).

PR #371, opened 2026-04-22
Remove stale Darwin hwprefs spec

tagliala:
The darwin10 example still stubs `hwprefs_available?`, but `processor_count` no longer calls that helper.

Commit 47253e9 added that fallback for old OS X 10.6 systems.

Commit 50bb18b rewrote processor detection and removed the helper, so this example now covers a dead path rather than supported Ruby 3.3+ behavior.


grosser (approved):
hehe nice find

PR #372, opened 2026-04-22
Prefer `String#include?` and `match?` over `=~`

tagliala:
Using `=~` to check for a plain substring is unnecessary: it allocates memory for `$~` and related global variables on every call, even when the result is only used as a boolean.

Prefer `String#include?` for plain string checks. When a regexp is needed (e.g. in specs), prefer `match?`. Neither pollutes the global regexp state.

Also prefer `match` over `=~` for more readable specs

Benchmarks:
```

Comparison (IPS):
            include?: 25395688.0 i/s
              match?: 19470868.8 i/s - 1.30x  (± 0.00) slower
                  =~: 10289823.3 i/s - 2.47x  (± 0.00) slower

Comparison (Memory):
              match?:          0 allocated
            include?:          0 allocated - same
                  =~:        168 allocated - Infx more
```

### DIFF
--- a/lib/parallel.rb
+++ b/lib/parallel.rb
@@ -266,7 +266,7 @@ module Parallel
 
       if options[:in_processes] && options[:in_threads]
         raise ArgumentError, "Please specify only one of `in_processes` or `in_threads`."
-      elsif RUBY_PLATFORM =~ /java/ && !options[:in_processes]
+      elsif RUBY_PLATFORM.include?('java') && !options[:in_processes]
         method = :in_threads
         size = options[method] || processor_count
       elsif options[:in_threads]
--- a/spec/parallel_spec.rb
+++ b/spec/parallel_spec.rb
@@ -261,12 +261,12 @@ describe Parallel do
     worker_types.each do |type|
       it "does not queue new work when one fails in #{type}" do
         out = `METHOD=map WORKER_TYPE=#{type} ruby spec/cases/with_exception.rb 2>&1`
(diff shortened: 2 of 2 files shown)


grosser (commented):
the java makes sense, but we don't need to optimize test runtime :)

tagliala:
Hi,

that is not for optimization

It's mentioned in the commit message

> Also prefer `match` over `=~` for more readable specs

```diff
- should ~=
+ should match
```

No problem in reverting the change btw

grosser:
hmmm true ... I like =~ more, but agree that match is easier to read and it's a good pattern to follow anyways

tagliala:
Thanks

I noticed that there is production code related to JRuby, but JRuby itself is not currently tested in the CI pipeline.

I attempted to run the test suite locally using JRuby, but it appears not to be fully compatible at the moment, primarily due to the lack of support for `fork`, which JRuby does not implement.

I also experimented with a potential workaround using Copilot, but the result wasn’t entirely satisfactory and would likely need a more thoughtful approach.

Would you be open to considering adding JRuby to the CI matrix as a starting point? The suite is not fully passing yet, having visibility into failures could help guide incremental improvements and better understand the current gaps in compatibility.

If this sounds reasonable, I’d be happy to explore potential changes or contribute toward making the test suite more JRuby-friendly.

Thanks for your thoughts!

grosser:
yeah fine to add jruby and just make it skip anything with fork for example

